PR #3, opened 2022-03-04
Add header of the home screen

adrifdez:
Add header design to project


src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "&:hover": {
+    backgroundColor: theme.palette.grey.light,
+  },
+});
+
+const useHeaderHomeStyles = createUseStyles((theme: Theme) => ({
+  header: {
+    ...flexBox,
+    justifyContent: "space-between",
```
amonsosanz, line 30: Why do we need the `justifyContent: "space-between"` and the `width: "100%"`?

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+    fontWeight: theme.fontWeights.medium,
+    "& > *": {
+      width: "100%",
```
amonsosanz, line 33: Why is this required? I mean, the logo and the subtitle will naturally get all the space they need.

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+  },
+  settingsBtn: {
+    ...flexBox,
+    alignItems: "center",
+    "& a": {
+      ...btn(theme),
+    },
+  },
+  myAccountBtn: {
+    ...flexBox,
```
amonsosanz, line 44: Given that these buttons are in a flex parent, instead of the `width: "100%"` I'd use `flex: 1`.

src/views/home/home.styles.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { Theme } from "src/styles/theme";
+
+const useHomeStyles = createUseStyles((theme: Theme) => ({
+  chainRow: {
+    textAlign: "center",
+  },
+  chain: {
+    display: "inline-flex",
+    justifyContent: "center",
```
amonsosanz, line 11: Is this required?

src/views/shared/container/container.styles.ts
```diff
@@ -0,0 +1,24 @@
...
+  root: {
+    width: "100%",
+    justifyContent: "center",
+    display: "flex",
+    backgroundColor: ({ background }: ContainerStyles) =>
+      background ? theme.palette.white : theme.palette.primary.main,
+  },
+  wrapper: {
+    width: "100%",
+    maxWidth: 700 + 2 * theme.spacing(3.5),
```
amonsosanz, line 19: Why this computation?

src/views/home/components/header/header.styles.ts
adrifdez: True I forgot remove it when I redesigned it

src/views/home/components/header/header.styles.ts
adrifdez: it's need it to keep the logo in the real center with the Bridge text

src/views/home/home.styles.ts
adrifdez: No because it's align it by the padding

src/views/shared/container/container.styles.ts
adrifdez: I brought this from `wallet-ui` because I didn't find any reference to the wrapper

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+  header: {
+    ...flexBox,
+    justifyContent: "space-between",
+    fontWeight: theme.fontWeights.medium,
+    "& > *": {
+      width: "100%",
+    },
+  },
+  settingsBtn: {
+    ...flexBox,
```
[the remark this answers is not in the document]
adrifdez: I prefer this way than create a new class and add two classes to `className` attribute but I'm open to any solution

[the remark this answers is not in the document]
amonsosanz: Let's wait for @elias-garcia PoV

src/styles/theme.ts
```diff
@@ -17,12 +17,12 @@ const theme = {
       hover: "#000411",
     },
     grey: {
-      veryLight: "#e1e1f1",
-      light: "#f3f3f8",
+      veryLight: "#f3f3f8",
```
elias-garcia, line 20: I would remove all the colors that we are not using in the bridge by the moment. Apart from that I would try to put some order in this file.

src/styles/theme.ts
```diff
@@ -17,12 +17,12 @@ const theme = {
...
+      light: "#E2E5EE",
```
elias-garcia, line 21: Use lowercase for the hex colors to be consistent please.

src/routing/routes.tsx
```diff
@@ -0,0 +1,13 @@
+import { Routes, Route } from "react-router-dom";
+
+import Home from "src/views/home/home.view";
+
+function Router(): JSX.Element {
```
elias-garcia, line 5: The file is named `routes.tsx` but the component is called `Router`, which is the correct one? `Router` looks better.

elias-garcia:
The icon `ethIcon` should be renamed to `eth-icon.svg`, we don't use camel case for filenames.

src/views/home/components/header/header.styles.ts
amonsosanz: But in this case, we don't need `flex: 1`, so instead of the spread I would add `display: "flex"`.

src/views/home/components/header/header.styles.ts
amonsosanz: I may not get the point but I can't appreciate any difference:

bridge-logo-centering

src/views/shared/container/container.styles.ts
amonsosanz: > I didn't find any reference to the wrapper

I don't know what you mean by this.

The point is, if we're doing this computation, there must be a reason. I wonder why can't we just set `max-width: 756px;`.

src/views/home/components/header/header.styles.ts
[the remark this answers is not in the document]
elias-garcia: I agree with Alex here. The way to abstract things is creating components, but not creating this style constants. If you want to abstract a button for example, just create a `Button` component. But I wouldn't do it by the moment, I would just inline the styles.

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,64 @@
...
+  "&:hover": {
+    backgroundColor: theme.palette.grey.light,
+  },
+});
+
+const useHeaderHomeStyles = createUseStyles((theme: Theme) => ({
+  header: {
+    ...flexBox,
+    fontWeight: theme.fontWeights.medium,
+    "& > *": {
```
elias-garcia, line 31: This kind of wildcards are very dangerous. Just apply the styles to the children that need it.

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,64 @@
...
+      flex: 1,
+    },
+  },
+  settingsBtn: {
```
elias-garcia, line 35: Avoid this kind of abbreviations, `settingsButton` is more descriptive and readable.

src/views/home/components/transaction-actions/transaction-actions.styles.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { Theme } from "src/styles/theme";
+
+const useTransactionActionsStyles = createUseStyles((theme: Theme) => ({
+  buttons: {
+    display: "flex",
+    justifyContent: "center",
+    margin: [[theme.spacing(1), theme.spacing(5)]],
+  },
+  actionBtn: {
```
elias-garcia, line 11: Same as before, use `actionButton` please

src/views/home/components/transaction-actions/transaction-actions.view.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ReactComponent as Deposit } from "src/assets/icons/deposit.svg";
+import { ReactComponent as Withdraw } from "src/assets/icons/withdraw.svg";
+import useTransactionActionsStyles from "src/views/home/components/transaction-actions/transaction-actions.styles";
+
+const TransactionActions = () => {
+  const classes = useTransactionActionsStyles();
+  return (
+    <div className={classes.buttons}>
+      <Link to={"/"} className={classes.actionBtn}>
+        <p className={classes.circle}>
```
elias-garcia, line 12: Why the `circle` is a `<p>` element?

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,64 @@
...
+    ...flexBox,
+    alignItems: "center",
+    "& a": {
+      ...btn(theme),
+    },
+  },
+  myAccountBtn: {
```
elias-garcia, line 42: `myAccountButton`

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,64 @@
...
+    },
+  },
+  logo: {
+    alignItems: "center",
+    textAlign: "center",
+    flexDirection: "column",
+    "& svg": {
+      width: 43,
+    },
+    "& p": {
```
elias-garcia, line 57: Avoid targeting inner elements, what if we change from a `p` to an `h1`? It's always better to have an specific class for each element, like `logoText`. Far more robust to changes.

elias-garcia (changes requested):

src/views/shared/container/container.styles.ts
adrifdez: Ahh to be equal than the padding

elias-garcia: This was a tricky thing from the wallet... I can't remember why but it has a reason. But we can remove it from here and set it as Alex said.

src/views/home/components/header/header.styles.ts
adrifdez: Peek 2022-03-04 13-44

src/views/home/components/transaction-actions/transaction-actions.view.tsx
adrifdez: because Link is an `a` element

src/views/home/components/header/header.styles.ts
amonsosanz: Aha! it's a browser-specific issue!

amonsosanz (approved):

src/views/home/components/header/header.styles.ts
```diff
@@ -0,0 +1,56 @@
+import { createUseStyles } from "react-jss";
+
+import { Theme } from "src/styles/theme";
+
+const useHeaderHomeStyles = createUseStyles((theme: Theme) => ({
```
elias-garcia, line 5: This variable should be called `useHeaderStyles`

src/views/home/components/transaction-actions/transaction-actions.styles.ts
elias-garcia: I still see `actionButton` here

src/views/home/components/header/header.view.tsx
```diff
@@ -0,0 +1,32 @@
+import { Link } from "react-router-dom";
+
+import { ReactComponent as Logo } from "src/assets/hermez-logo.svg";
+import { ReactComponent as Settings } from "src/assets/icons/settings.svg";
```
elias-garcia, line 4: When you import an icon it would be great to add the `Icon` prefix:

```
import { ReactComponent as SettingsIcon } from "src/assets/icons/settings.svg";
import { ReactComponent as QRIcon } from "src/assets/icons/qr.svg";
```

src/styles/theme.ts
```diff
@@ -2,57 +2,25 @@ const theme = {
...
-      main: "#8248e5",
-      dark: "#6d00f1",
+      light: "#e2e5ee",
+      hover: "rgba(67, 85, 140, 0.1)",
     },
-    green: "#219653",
   },
   hoverTransition: "all 100ms",
   fontWeights: {
-    normal: "400",
```
elias-garcia, line 45: We can keep the `fontWeights` map as we are going to need it. At least I would keep this:

```
fontWeights: {
    normal: "400",
    medium: "500",
    bold: "700",
}
```

src/views/home/components/transaction-actions/transaction-actions.view.tsx
```diff
@@ -0,0 +1,27 @@
+import { Link } from "react-router-dom";
+
+import { ReactComponent as Deposit } from "src/assets/icons/deposit.svg";
```
elias-garcia, line 3:
```ts
import { ReactComponent as DepositIcon } from "src/assets/icons/deposit.svg";
import { ReactComponent as WithdrawIcon } from "src/assets/icons/withdraw.svg";
```

elias-garcia (changes requested):

src/views/home/components/transaction-actions/transaction-actions.view.tsx
elias-garcia: I know, but a <p> element is for paragraphs. Why you don't apply the circle styles to the `a` element and you keep just the icon inside?

src/styles/theme.ts
adrifdez: I'll add just bold because normal is by default

elias-garcia (approved):

PR #5, opened 2022-03-09
Add Activity Screen

adrifdez:
### What does this PR does?

Add Activity Screen

## Checklist

review them:


src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+import FiatAmount from "src/views/shared/fiatAmount/fiat-amount";
+import Typography from "src/views/shared/typography/typography.view";
+import Card from "src/views/shared/card/card.view";
+
+export interface TransactionCardProps {
+  type: "l1" | "l2";
+  id: number;
+  timestamp: number;
+  token: "eth" | "dai";
+  status: "hold" | "initiate" | "process" | "complete" | "error";
```
elias-garcia, line 16: `status: "processing" | "initiated" | "on-hold" | "completed" | "failed" `

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+  amount: number;
+}
+
+const TrasnsactionCard = ({ type, timestamp, token, amount, status }: TransactionCardProps) => {
```
elias-garcia, line 20: Typo `TrasnsactionCard` -> TransactionCard

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const classes = useTransactionCardStyles();
+  const IconsLayer = {
+    l1: TransferL1Icon,
+    l2: TransferL2Icon,
+  };
+  const Icon = status !== "complete" && status !== "error" ? ReloadIcon : IconsLayer[type];
```
elias-garcia, line 26: We need to create a shared component called `Spinner` to represent the spinning effect, as it needs to spin. I wouldn't add the animation by the moment as Chami asked for a very specific animation, but I would create the component and move this condition to the JSX.

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+              }`}
+            >
+              {statusText[status]}
+            </span>
+          )}
+        </div>
+        <div className={classes.tokenColumn}>
+          <div className={classes.token}>
+            <TokenIcon className={classes.tokenIcon} />
+            <Typography type={"body1"}>
```
elias-garcia, line 85: Brackets are not needed in multiple parts of this file as well.

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+          </Typography>
+        </div>
+      </div>
+      {status === "initiate" && (
+        <div className={classes.bottom}>
+          <Typography type={"body2"}>
+            Step 2 will require signature in {convertToTime()}.
+          </Typography>
+          <button disabled className={classes.finaliseButton}>
+            Finalise{" "}
```
elias-garcia, line 100: Why we need this empty space after `Finalise`?

src/views/activity/activity.view.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const Activity = () => {
+  const classes = useActivityStyles();
+  const [listAll, setListAll] = useState(true);
+  const pendingActivities = demoData.filter(
+    (data) => data.status !== "complete" && data.status !== "error"
+  );
+  const transactionsList = listAll ? demoData : pendingActivities;
+
+  const changeList = () => {
```
elias-garcia, line 17: Handlers should be prefixed with `on` -> `onChangeList`

src/views/activity/activity.view.tsx
```diff
@@ -0,0 +1,61 @@
...
+    setListAll(!listAll);
+  };
+
+  return (
+    <>
+      <Header title={"Activity"} />
+      <div className={classes.selectorBoxes}>
+        <div
+          className={`${classes.selectorBox} ${!listAll ? classes.inactiveBox : ""}`}
+          onClick={!listAll ? changeList : undefined}
```
elias-garcia, line 27: It's not a good practice to add or remove handlers conditionally:

```
const onDisplayAll = () => setListAll(true)
const onDisplayPending = () => setListAll(false)

onClick={onDisplayAll}
onClick={onDisplayPending}
```

Doing this we avoid having conditional handlers and the component state won't change if you click more than once a button, so it's not going to trigger a new render.

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const classes = useTransactionCardStyles();
+  const IconsLayer = {
+    l1: TransferL1Icon,
+    l2: TransferL2Icon,
+  };
+  const Icon = status !== "complete" && status !== "error" ? ReloadIcon : IconsLayer[type];
+  const actionText = type === "l1" ? "Transfer to L1" : "Transfer to L2";
+  const statusText = {
+    hold: "On Hold",
+    initiate: "Initiate",
```
amonsosanz: As the types, all the literals should be a participle.

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,112 @@
...
+    process: "Processing",
+    error: "Error",
```
amonsosanz: Don't we need a status text for `completed`?

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+  };
+  const tokenIcons = {
+    eth: EthToken,
+    dai: DaiToken,
+  };
+  const TokenIcon = tokenIcons[token] || tokenIcons["eth"];
+
+  const convertToTime = (): string => {
```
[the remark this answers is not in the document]
adrifdez: I think that it's more an util than a component

src/views/activity/activity.view.tsx
```diff
@@ -0,0 +1,61 @@
...
+import TrasnsactionCard from "src/views/activity/components/transaction-card/transaction-card.view";
+import useActivityStyles from "src/views/activity/activity.styles";
+import Typography from "src/views/shared/typography/typography.view";
+import Header from "src/views/shared/header/header.view";
+import { demoData } from "src/views/activity/demo-data";
+
+const Activity = () => {
+  const classes = useActivityStyles();
+  const [listAll, setListAll] = useState(true);
+  const pendingActivities = demoData.filter(
```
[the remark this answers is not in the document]
amonsosanz: Or a `reduce`, instead of a `filter` +  a `map`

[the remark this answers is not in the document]
adrifdez: Because we need to count pending transactions

src/views/activity/components/transaction-card/transaction-card.view.tsx
adrifdez: But one thing it's the component and other the logic that I think that should be here because the shared spinner could have differents status that these ones. 
While we don't have the animation I'm not see necessary separate it

src/views/activity/components/transaction-card/transaction-card.view.tsx
adrifdez: We don't show the tag when is `completed`

src/views/activity/activity.view.tsx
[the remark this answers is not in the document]
amonsosanz: Ah right now I see the `pendingActivities.length`

amonsosanz (approved):

src/views/activity/components/transaction-card/transaction-card.view.tsx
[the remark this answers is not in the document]
elias-garcia: Nice, I agree that it's more a util than a component. But then I'll follow the same approach for the `FiatAmount` component, as it returns `<>{fiatAmount}</>`. I'll follow the same pattern for this.

Btw, I'll rename the `time-ago.ts` file to `time.ts`. Remember that module names should be more generic, and specific functions like `timeAgo` should be inside. I also suggest to use named exports for this util files instead of a default export.

src/views/activity/activity.view.tsx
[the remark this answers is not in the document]
elias-garcia: Ahh nice!  I didn't see that

src/utils/amounts.ts
```diff
@@ -0,0 +1,14 @@
...
+  amount: number;
+  token: "eth" | "dai";
+}
+
+const rates = {
+  dai: 1.02,
+  eth: 2500,
+};
+
+export const fiatAmount = ({ amount, token }: FiatAmountProps): string => {
```
elias-garcia, line 11: Remember that as now it's a function and not a component it should be a verb (not a name). Some options could be `convertTokenAmountToFiat` or `getFiatAmount`.

Also, take into account that you are using one object as a param, but in the `timeAgo` function you are using just params. I'll choose one approach and apply the same to all the utils. If you stick with this approach, remember the `FiatAmountProps` to `FiatAmountParams` as it makes more sense, `props` keyword should be used for React components.

src/utils/time.ts
```diff
@@ -0,0 +1,17 @@
+export const timeAgo = (timestamp: number): string => {
```
elias-garcia, line 1: Same as before, remember that functions should be verbs: `const getTimeFromNow = ...`

elias-garcia (approved):

PR #6, opened 2022-03-11
Add Settings screen

adrifdez:
### What does this PR does?

Add Setting screen

## Checklist

review them:


amonsosanz:
There's an issue with the break-points:

settings-currency-jump

src/views/settings/settings.view.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <div className={classes.currencies}>
+            {currencies.map((currency) => (
+              <div
+                className={`${classes.currencyBox} ${
+                  currency.id === currencySelected ? classes.currencySelected : ""
+                }`}
+                onClick={() => setCurrencySelected(currency.id)}
+                key={currency.id}
+              >
+                {currency.id === currencySelected ? <CheckedIcon /> : <UncheckedIcon />}{" "}
```
amonsosanz, line 52: What's this `{" "}` for?

src/views/settings/settings.view.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <>
+      <Header title="Settings" />
+      <Typography type="body2" className={classes.subtitle}>
+        Hermez Polygon Bridge v.01
+      </Typography>
+      <Card className={classes.card}>
+        <div className={classes.currenciesRow}>
+          <div>
```
amonsosanz, line 36: Why is this div required?

src/views/settings/settings.styles.ts
```diff
@@ -0,0 +1,74 @@
+import { createUseStyles } from "react-jss";
+
+import { Theme } from "src/styles/theme";
+
+const useActivityStyles = createUseStyles((theme: Theme) => ({
+  subtitle: {
+    marginTop: theme.spacing(-2),
```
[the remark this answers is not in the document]
adrifdez: I thought that at the beginning but this subtitle just exist in this screen for that I leave it there while we don't have more subtitles in the app

src/views/settings/settings.view.tsx
adrifdez: Maybe from previous design

src/views/settings/settings.styles.ts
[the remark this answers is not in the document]
amonsosanz: Fair enough, let's move it to the Header when we need it a second time.

src/views/settings/settings.view.tsx
amonsosanz: Prettier is adding this `{" "}`?? Wow.

amonsosanz (approved):

src/views/settings/settings.view.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { ReactComponent as LogoutIcon } from "src/assets/icons/logout.svg";
+import { ReactComponent as EurIcon } from "src/assets/currencies/eur.svg";
+import { ReactComponent as UsdIcon } from "src/assets/currencies/usd.svg";
+import { ReactComponent as JpyIcon } from "src/assets/currencies/jpy.svg";
+import { ReactComponent as GbpIcon } from "src/assets/currencies/gbp.svg";
+import { ReactComponent as CnyIcon } from "src/assets/currencies/cny.svg";
+
+const Settings = () => {
+  const classes = useSettingsStyles();
+  const [currencySelected, setCurrencySelected] = useState("eur");
```
elias-garcia, line 19: `const [selectedCurrencyId, setSelectedCurrencyId] = useState("eur");`

elias-garcia: BTW, I think that the type of `selectedCurrencyId` is `string`. We should create an enum or a type for all the currency ids: eur, usd, gbp, cny and jpy and make `selectedCurrencyId` use this type instead of an string.

src/views/settings/settings.view.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <div className={classes.currencies}>
+            {currencies.map((currency) => (
+              <div
+                className={`${classes.currencyBox} ${
+                  currency.id === currencySelected ? classes.currencySelected : ""
+                }`}
+                onClick={() => setCurrencySelected(currency.id)}
+                key={currency.id}
+              >
+                {currency.id === currencySelected ? <CheckedIcon /> : <UncheckedIcon />}
```
elias-garcia, line 50: This list should be a radio group: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio.

src/views/settings/settings.view.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { ReactComponent as UncheckedIcon } from "src/assets/icons/checkbox-unchecked.svg";
+import { ReactComponent as LogoutIcon } from "src/assets/icons/logout.svg";
+import { ReactComponent as EurIcon } from "src/assets/icons/currencies/eur.svg";
+import { ReactComponent as UsdIcon } from "src/assets/icons/currencies/usd.svg";
+import { ReactComponent as JpyIcon } from "src/assets/icons/currencies/jpy.svg";
+import { ReactComponent as GbpIcon } from "src/assets/icons/currencies/gbp.svg";
+import { ReactComponent as CnyIcon } from "src/assets/icons/currencies/cny.svg";
+
+enum Currencies {
+  eur = "eur",
```
elias-garcia, line 18: TypeScript `enum` keys should be in mayus:

```ts
enum Currencies {
  EUR = "eur",
  // ...
```

src/views/settings/settings.view.tsx
```diff
@@ -0,0 +1,82 @@
...
+          <Typography type="body1" className={classes.row}>
+            <ConversionIcon className={classes.conversionIcon} />
+            Currency conversion
+          </Typography>
+          <Typography type="body2">Select a currency for conversion display</Typography>
+          <div className={classes.currencies}>
+            {currencies.map((currency) => (
+              <label
+                className={`${classes.currencyBox} ${
+                  currency.id === selectedCurrencyId ? classes.currencySelected : ""
```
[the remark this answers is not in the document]
adrifdez: The label is not getting the `:checked` status from the input, I can use some selector code in the style but normally we don't want it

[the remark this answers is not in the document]
elias-garcia: Mmmm I see! It's perfect then, good job :)

One problem that I've found testing it is that the `cursor: pointer` is not working when you are hovering the checkbox.

elias-garcia (approved):

PR #7, opened 2022-03-14
Add support for login with MetaMask + WalletConnect

elias-garcia:
### What does this PR does?

This PR adds support to login into the dapp using MetaMask or WalletConnect. It also handles `EIP-1193` events and acts accordingly. This is just a first iteration, we need to continue working specially on error handling.

`@walletconnect/ethereum-provider` has been forced to `1.7.1` version as an uncaught error is happening when closing the WalletConnect modal in the latest release: https://github.com/WalletConnect/walletconnect-monorepo/issues/839

### How to test?

Login with MetaMask and with WalletConnect.

## Checklist

review them:

### DIFF
--- a/src/domain/index.ts
+++ b/src/domain/index.ts
@@ -1,6 +1,6 @@
 export enum WalletName {
-  METAMASK,
-  WALLET_CONNECT,
+  METAMASK = "MetaMask",
+  WALLET_CONNECT = "WalletConnect",
 }
 
 export enum EthereumEvent {
--- a/src/views/login/components/account-loader/account-loader.view.tsx
+++ b/src/views/login/components/account-loader/account-loader.view.tsx
@@ -5,11 +5,6 @@ import { WalletName } from "src/domain";
 import Typography from "src/views/shared/typography/typography.view";
 import WalletIcon from "../wallet-icon/wallet-icon.view";
 
-const walletLabels = {
-  [WalletName.METAMASK]: "MetaMask",
-  [WalletName.WALLET_CONNECT]: "WalletConnect",
-};
-
 interface AccountLoaderProps {
   selectedWallet: WalletName;
 }
@@ -21,10 +16,10 @@ const AccountLoader: FC<AccountLoaderProps> = ({ selectedWallet }) => {
     <div className={classes.accountLoader}>
       <WalletIcon walletName={selectedWallet} size="lg" className={classes.walletIcon} />
       <Typography type="body1" className={classes.walletName}>
-        {walletLabels[selectedWallet]}
+        {selectedWallet}
       </Typography>
       <Typography type="body2" className={classes.hint}>
-        Follow the instructions in {walletLabels[selectedWallet]}
+        Follow the instructions in {selectedWallet}
       </Typography>
     </div>
   );
--- a/src/views/login/components/wallet-list/wallet-list.view.tsx
+++ b/src/views/login/components/wallet-list/wallet-list.view.tsx
@@ -15,38 +15,27 @@ const WalletList: FC<WalletListProps> = ({ onSelectWallet }) => {
 
   return (
     <ul className={classes.walletList}>
-      <li
-        className={classes.wallet}
-        role="button"
-        onClick={() => onSelectWallet(WalletName.METAMASK)}
-      >
-        <WalletIcon walletName={WalletName.METAMASK} size="sm" className={classes.walletIcon} />
-        <div className={classes.walletInfo}>
-          <Typography type="body1" className={classes.walletName}>
-            MetaMask
-          </Typography>
-          <Typography type="body2">Connect using web wallet</Typography>
-        </div>
-        <CaretRightIcon />
-      </li>
-      <li
-        className={classes.wallet}
-        role="button"
-        onClick={() => onSelectWallet(WalletName.WALLET_CONNECT)}
-      >
-        <WalletIcon
-          walletName={WalletName.WALLET_CONNECT}
-          size="sm"
-          className={classes.walletIcon}
-        />
-        <div className={classes.walletInfo}>
-          <Typography type="body1" className={classes.walletName}>
-            WalletConnect
-          </Typography>
-          <Typography type="body2">Connect using mobile wallet</Typography>
-        </div>
-        <CaretRightIcon />
-      </li>
+      {Object.values(WalletName).map((walletName) => (
+        <li
+          key={walletName}
+          className={classes.wallet}
+          role="button"
+          onClick={() => onSelectWallet(walletName)}
+        >
+          <WalletIcon walletName={walletName} size="sm" className={classes.walletIcon} />
+          <div className={classes.walletInfo}>
+            <Typography type="body1" className={classes.walletName}>
+              {walletName}
+            </Typography>
+            {walletName === WalletName.METAMASK ? (
+              <Typography type="body2">Connect using web wallet</Typography>
+            ) : (
+              <Typography type="body2">Connect using mobile wallet</Typography>
+            )}
+          </div>
+          <CaretRightIcon />
+        </li>
+      ))}
     </ul>
   );
 };


src/contexts/ethereum-provider.context.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const navigate = useNavigate();
+  const [rawProvider, setRawProvider] = useState<MetaMaskProvider | WalletConnectProvider>();
+  const [provider, setProvider] = useState<Web3Provider>();
+  const [account, setAccount] = useState<AsyncTask<string, string>>({ status: "pending" });
+  const env = useEnvContext();
+
+  const connectProvider = async (walletName: WalletName): Promise<void> => {
+    switch (walletName) {
+      case WalletName.METAMASK: {
+        if (window.ethereum) {
```
amonsosanz, line 34: I think we should check the `ExternalProvider` to guarantee that it actually comes from Metamask. Like it's now we will accept a provider different from Metamask present on `window.ethereum`.

Regarding how to check it, at least, we should check for the prop `isMetaMask`.

A more ambitious approach would be to actually decode the object, which would let us declare the type `ethereum` from [`react-app-env.d.ts`](#diff-433083e9966b3a808a0b2dce3e47b0b74152c61cb7d5b6a4ad8aad2592dacd81R7) as `unknown` (what it actually is) until it's properly decoded and added to the context.

`ExternalProvider` is indeed a pretty simple interface and [zod allows us to parse objects with functions](https://github.com/colinhacks/zod#function-schemas).

src/contexts/ethereum-provider.context.tsx
```diff
@@ -0,0 +1,119 @@
...
+      rawProvider.on(EthereumEvent.ACCOUNTS_CHANGED, onAccountsChanged);
+      rawProvider.on(EthereumEvent.CHAIN_CHANGED, onChainChangedOrDisconnect);
+      rawProvider.on(EthereumEvent.DISCONNECT, onChainChangedOrDisconnect);
+    }
+
+    return () => {
+      if (rawProvider && rawProvider.removeListener) {
+        rawProvider.removeListener(EthereumEvent.ACCOUNTS_CHANGED, onAccountsChanged);
+        rawProvider.removeListener(EthereumEvent.CHAIN_CHANGED, onChainChangedOrDisconnect);
+        rawProvider.removeListener(EthereumEvent.CHAIN_CHANGED, onChainChangedOrDisconnect);
```
amonsosanz, line 105: It looks like here we should be using the instance of the class `Web3Provider` instead of the `ExternalProvider` object we pass to the constructor.

I guess we've had to create the interface `MetaMaskProvider` just to be able to manage the events here. There's another comment regarding that interface.

src/react-app-env.d.ts
```diff
@@ -1 +1,9 @@
 /// <reference types="react-scripts" />
+
+import { MetaMaskProvider } from "./domain";
+
+declare global {
+  interface Window {
+    ethereum?: MetaMaskProvider;
```
amonsosanz, line 7: This should be an `ExternalProvider` (and ideally, `unknown`).

src/contexts/ethereum-provider.context.tsx
amonsosanz: After our meeting, I understood the reasoning for the current implementation: `Web3Provider` instances do not support the [EIP-1193 events](https://eips.ethereum.org/EIPS/eip-1193#events) `on` and `removeListener`.

We could though reach the `ExternalProvider` present in the `Web3Provider` instance at `provider.provider` and make sure the methods we need exist:

```typescript
if (provider) {
  const rawProvider: Record<string, unknown> = provider.provider;
  if ("on" in rawProvider && typeof rawProvider.on === "function") {
    rawProvider.on(EthereumEvent.ACCOUNTS_CHANGED, onAccountsChanged);
    rawProvider.on(EthereumEvent.CHAIN_CHANGED, onChainChangedOrDisconnect);
    rawProvider.on(EthereumEvent.DISCONNECT, onChainChangedOrDisconnect);
  }
}

return () => {
  if (provider) {
    const rawProvider: Record<string, unknown> = provider.provider;
    if ("removeListener" in rawProvider && typeof rawProvider.removeListener === "function") {
      rawProvider.removeListener(EthereumEvent.ACCOUNTS_CHANGED, onAccountsChanged);
      rawProvider.removeListener(EthereumEvent.CHAIN_CHANGED, onChainChangedOrDisconnect);
      rawProvider.removeListener(EthereumEvent.DISCONNECT, onChainChangedOrDisconnect);
    }
  }
};
```

This approach is equally typesafe and has the advantage of limiting dirtiness to where we need to call the methods, in contrast to having to create the `MetaMaskProvider` interface and store the `rawProvider` in the context.

src/contexts/ethereum-provider.context.tsx
amonsosanz: I've given this a try and works beautifully.

In `src/react-app-env.d.ts` we set `ethereum` as `unknown`:

```typescript
/// <reference types="react-scripts" />

declare global {
  interface Window {
    ethereum: unknown; // at this point is all we know
  }
}

export {};
```

Later in `src/adapters/parsers.ts` we declare the parser of the provider, making sure that it's a valid **MetaMask** `ExternalProvider` (notice the `isMetaMask: z.boolean().refine((val) => val === true)`):

```typescript
const metamaskExternalProviderParser = StrictSchema<ExternalProvider>()(
  z
    .object({
      isMetaMask: z.boolean().refine((val) => val === true),
      isStatus: z.boolean(),
      host: z.string(),
      path: z.string(),
      sendAsync: z
        .function()
        .args(
          z.object({
            method: z.string(),
            params: z.array(z.any()).optional(),
          }),
          z.function().args(z.any(), z.any()).returns(z.void())
        )
        .returns(z.void()),
      send: z
        .function()
        .args(
          z.object({
            method: z.string(),
            params: z.array(z.any()).optional(),
          }),
          z.function().args(z.any(), z.any()).returns(z.void())
        )
        .returns(z.void()),
      request: z
        .function()
        .args(
          z.object({
            method: z.string(),
            params: z.array(z.any()).optional(),
          })
        )
        .returns(z.promise(z.void())),
    })
    .partial()
);
```

And finally in `src/contexts/ethereum-provider.context.tsx`, we decode the object and only proceed when it's a valid **MetaMask** `ExternalProvider`:

```typescript
const parsedMetamaskExternalProvider = metamaskExternalProviderParser.safeParse(window.ethereum);
if (parsedMetamaskExternalProvider.success) {...}
```

I experimented with this parser just for fun. I realize that going this way has the drawback of having to maintain the parser whenever the interface `ExternalProvider` is changed, so I'm actually not sure if this is the way to go.

At least though, we should tests for `window.ethereum && window.ethereum.isMetaMask` in the `if` statement.

Regarding this, I've confirmed that with the current implementation you can log in to the bridge using the [Coinbase Wallet extension](https://chrome.google.com/webstore/detail/coinbase-wallet-extension/hnfanknocfeofbddgcijnmhnfnkdnaad), an this may have undesired consequences.

src/adapters/parsers.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from "zod";
+
+import { StrictSchema } from "src/utils/type-safety";
+import { Env } from "src/domain";
+
+const envParser = StrictSchema<Env>()(z.object({ REACT_APP_INFURA_API_KEY: z.string() }));
+
+const ethereumAccountsParser = StrictSchema<string[]>()(z.array(z.string()));
```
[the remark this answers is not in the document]
elias-garcia: I don't agree with that approach. Because, if you need to use the parser more than one time you'll always need to write the `z.array()`... Also if the list of entities is not an array the parser could be different... Another thing that I don't like is to import `zod` outside of this module.

On what I agree is that we should have one parser for an entity and another one for the entity list (if needed), like this:

```ts
const ethereumAccountParser = StrictSchema<string>()(z.string());
```

But yes, in this case as we are not using the single entity and it's very straightforward, I agree that it's not necessary.

src/contexts/ethereum-provider.context.tsx
elias-garcia: Done!

src/contexts/ethereum-provider.context.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+interface EthereumProviderContextData {
+  provider?: Web3Provider;
+  account: AsyncTask<string, string>;
+  connectProvider?: (walletName: WalletName) => Promise<void>;
+}
+
+const EthereumProviderContext = createContext<EthereumProviderContextData>({
+  provider: undefined,
+  account: { status: "pending" },
```
amonsosanz, line 21: Why are we explicit with `provider: undefined` and not with `connectProvider: undefined`?

src/contexts/ethereum-provider.context.tsx
```diff
@@ -0,0 +1,120 @@
...
+          return getConnectedAccounts(web3Provider)
+            .then((accounts) => {
+              setProvider(web3Provider);
+              setAccount({ status: "successful", data: accounts[0] });
+            })
+            .catch(() => setAccount({ status: "failed", error: "Error loading account" }));
+        } else {
+          return setAccount({
+            status: "failed",
+            error: "There is no Ethereum provider injected in the window",
```
amonsosanz, line 46: Wouldn't it make sense to have a meaningful error that we can show to the user? Something in the line of:

> We cannot detect your wallet. Make sure the MetaMask extension is installed and active in your browser.

src/views/home/home.view.tsx
```diff
@@ -2,16 +2,21 @@ import useHomeStyles from "src/views/home/home.styles";
...
   const classes = useHomeStyles();
+  const { account } = useEthereumProviderContext();
 
   return (
     <>
       <Header />
       <div className={classes.ethereumAddress}>
         <MetaMaskIcon className={classes.metaMaskIcon} />
-        <Typography type="body1">0x2387 ･･･ 5682</Typography>
+        {account.status === "successful" && (
```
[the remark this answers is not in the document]
elias-garcia: Good catch! Fixed

src/contexts/ethereum-provider.context.tsx
elias-garcia: Good catch! Fixed

src/react-app-env.d.ts
elias-garcia: Moved to a `ExternalProvider` by the moment

src/contexts/ethereum-provider.context.tsx
elias-garcia: Yes, this is going to be done in the first PR. I still need to figure out how are we going to manage the `chainId` as we need to take into account also if we are on mainnet environment or testnet environment.

src/contexts/ethereum-provider.context.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { useNavigate } from "react-router-dom";
+
+import { EthereumEvent, WalletName } from "src/domain";
+import { AsyncTask } from "src/utils/types";
+import { ethereumAccountsParser } from "src/adapters/parsers";
+import { getConnectedAccounts } from "src/adapters/ethereum";
+import { useEnvContext } from "./env.context";
+import routes from "src/routes";
+
+interface EthereumProviderContextData {
```
amonsosanz, line 13: I'd remove this `Data` suffix.

amonsosanz (approved):

adrifdez (approved):

PR #9, opened 2022-03-17
Error management

amonsosanz:
This PR implements the management of the errors, mainly by:

 * Parsing every caught error and:
   * Showing component-located, red, informative errors to the user.
   * Showing floating, snackbar-based errors to the user.
     * Including the ability to report errors through a Google Form.


src/adapters/error.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export const REPORT_ERROR_FORM_ENTRIES = {
+  url: "entry.2056392454",
+  network: "entry.1632331664",
+  platform: "entry.259085709",
+  error: "entry.1383309652",
+};
+
+export const REPORT_ERROR_FORM_URL =
+  "https://docs.google.com/forms/d/1YOvhK2RfTQmYO8DGMRqN7FYxRhBZd9jB6PZ7InJirTk/viewform";
```
amonsosanz, line 95: Should we add a `constants` module for these? Or maybe use env vars?

amonsosanz, line 95: Should we add a `constants` module for these? Or maybe use env vars?

src/adapters/error.ts
```diff
@@ -0,0 +1,109 @@
+import { z, ZodError } from "zod";
+import * as platform from "platform";
+import * as StackTrace from "stacktrace-js";
+
+import { StrictSchema } from "src/utils/type-safety";
+
+interface MessageKeyError {
```
elias-garcia, line 7: Should we add this interfaces to our domain?

src/adapters/error.ts
elias-garcia: Yep, I would create a `constants` module for `REPORT_ERROR_FORM_ENTRIES` and for `REPORT_ERROR_FORM_URL`

src/assets/icons/close-white.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="30" height="30" fill="none" viewBox="0 0 30 30">
```
elias-garcia, line 1: I would rename this icon to `xmark-white`, as we already have one with this name

src/views/login/login.view.tsx
```diff
@@ -40,23 +40,29 @@ const Login: FC = () => {
...
+                type="h1"
+                className={`${classes.cardHeader} ${classes.cardHeaderCentered}`}
+              >
+                Connecting to
+              </Typography>
+              <AccountLoader selectedWallet={selectedWallet} />
+            </>
+          )}
+        </Card>
+        {account.status === "failed" && <Error error={account.error} />}
```
elias-garcia, line 64: I think that this error should disappear when you click again in a wallet to login. I think that this should work:

```
{selectedWallet === undefined && account.status === "failed" && <Error error={account.error} />}
```

src/contexts/ethereum-provider.context.tsx
```diff
@@ -60,9 +73,18 @@ const EthereumProviderProvider: FC = (props) => {
               setProvider(web3Provider);
               setAccount({ status: "successful", data: accounts[0] });
             })
-            .catch(() => {
-              setAccount({ status: "failed", error: "Error loading account" });
-            });
+            .catch((error) =>
+              parseError(error).then((errorMsg) => {
+                if (error instanceof Error && error.message === "User closed modal") {
```
amonsosanz, line 78: Not very proud of this, but how can we otherwise identify a "User closed modal" error from WalletConnect?

src/contexts/global.context.tsx
```diff
@@ -0,0 +1,59 @@
...
+const globalContextNotReadyErrorMsg = "The global context is not yet ready";
+
+const globalContextDefaultValue: GlobalContext = {
+  snackbar: { status: "closed" },
+  openSnackbar: () => {
+    console.error(globalContextNotReadyErrorMsg);
+  },
+  closeSnackbar: () => {
+    console.error(globalContextNotReadyErrorMsg);
+  },
```
amonsosanz, line 29: I did this to prevent declaring `openSnackbar` and `closeSnackbar` as optional. `openSnackbar` will be used many times and we don't want to check for `undefined` every single time.

src/contexts/ethereum-provider.context.tsx
```diff
@@ -26,24 +28,35 @@ const EthereumProviderProvider: FC = (props) => {
...
           return getConnectedAccounts(web3Provider)
             .then((accounts) => {
               setProvider(web3Provider);
               setAccount({ status: "successful", data: accounts[0] });
             })
-            .catch(() => setAccount({ status: "failed", error: "Error loading account" }));
+            .catch((error) =>
+              parseError(error).then((errorMsg) => {
+                if (isMetamaskUserRejectedRequestError(error)) {
+                  setAccount({ status: "failed", error: errorMsg });
```
elias-garcia, line 47: In this case I wouldn't display any error, as it's not relevant for the user

src/contexts/global.context.tsx
```diff
@@ -0,0 +1,59 @@
...
+      message: Message;
+    };
+
+interface GlobalContext {
+  snackbar: SnackbarState;
+  openSnackbar: (message: Message) => void;
+  closeSnackbar: () => void;
+}
+
+const globalContextNotReadyErrorMsg = "The global context is not yet ready";
```
elias-garcia, line 20: I like this!

src/views/core/layout/layout.view.tsx
```diff
@@ -1,14 +1,30 @@
...
+import Snackbar from "src/views/shared/snackbar/snackbar.view";
+import { useGlobalContext } from "src/contexts/global.context";
+import { reportError } from "src/adapters/error";
 
 const Layout: FC = ({ children }) => {
   const classes = useLayoutStyles();
+  const { snackbar, closeSnackbar } = useGlobalContext();
+
+  const onCloseSnackbar = closeSnackbar;
+  const onReportFromSnackbar = reportError;
```
amonsosanz, line 13: I don't feel the need to do this, but I wanted to avoid the PR comment about the missing `on` prefix on the handlers.

We can go 3 ways regarding this:

 1. Reassigning the handlers as it's now.
 1. Naming the handlers with the `on` prefix at the source (GlobalContext)
 1. Accepting `onClose={closeSnackbar}` and `onReport={reportError}`

I'm ok with the three.

src/views/login/login.view.tsx
amonsosanz: It should be already disappearing since `account` goes back to `{status: "loading"}`

src/views/shared/snackbar/snackbar.view.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+
+import { ReactComponent as ErrorIcon } from "src/assets/icons/error.svg";
+import { ReactComponent as CloseIconLight } from "src/assets/icons/close-white.svg";
+import useSnackbarStyles from "src/views/shared/snackbar/snackbar.styles";
+//domain
+import { Message } from "src/domain";
+
+const SNACKBAR_AUTO_HIDE_DURATION = 5000;
```
amonsosanz, line 9: Movi this to `constants`.

src/contexts/global.context.tsx
amonsosanz: I did it to prevent declaring `openSnackbar` and `closeSnackbar` as optional. `openSnackbar` will be used many times and we don't want to check for `undefined` every single time.

src/adapters/error.ts
amonsosanz: Done

src/adapters/error.ts
amonsosanz: We have discussed this topic in the past for other types. This is the definition of the Domain that I wrote [some time ago](https://www.notion.so/iden3/Wallet-UI-Typescript-refactor-f8bd6c2fe6db46ffb40ff2419441fed6#5c35d68155cd43d1b42d5b1d574247d6):

> The domain layer contains the abstract definition of the types that are used across the application. These types should generally correspond to real entities of the domain of the application. Special care must be taken on the decision of what constitutes a domain entity, as these types will determine the general discourse of the code of the whole application. Think of it as the glossary of the application.

So no, I don't think we should add `MessageKeyError` to the domain since it's not a central type of the application. It's just an accessory type used by an error parser in `adapters`.

src/assets/icons/close-white.svg
amonsosanz: mmm... I was not aware of this icon. We should make them the same!

image

Which one you like the most?

amonsosanz: Done as agreed. Removed the `close-white` icon at the left and added a white version of the existing `xmark` icon.

src/contexts/ethereum-provider.context.tsx
amonsosanz: Right. Done.

src/views/shared/snackbar/snackbar.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+  reportButton: {
+    textTransform: "uppercase",
+    color: theme.palette.white,
+    border: "1px solid",
+    backgroundColor: "transparent",
+    margin: `0 ${theme.spacing(1)}px`,
+    lineHeight: "14px",
+    padding: theme.spacing(1),
+    cursor: "pointer",
+    "&:hover:not(:disabled)": {
```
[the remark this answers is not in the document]
amonsosanz: Good catch! This is a leftover from the Wallet. In the Wallet, the [Report button was using a Button component](https://github.com/hermeznetwork/wallet-ui/blob/main/src/views/shared/snackbar/snackbar.view.tsx#L46) that was setting this attribute. Here this does not make any sense.

Removed.

adrifdez (approved):

src/assets/icons/close-white.svg
elias-garcia: But instead of adding the same versión why you don't use the black one and override the fill property of the path?

elias-garcia (approved):

PR #10, opened 2022-03-17
Add details screen

adrifdez:
### What does this PR does?

Add Transaction details Screen with demo data

## Checklist

review them:


src/views/transaction-details/components/chain-row/chain-row.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const chain = {
+    l1: { from: "hermez", to: "eth" },
+    l2: { from: "eth", to: "hermez" },
+  };
+
+  switch (chain[type][side]) {
+    case "eth":
+      return (
+        <>
```
elias-garcia, line 23: I think that we don't need to return this `<></>` wrappers.

src/views/transaction-details/components/chain-row/chain-row.tsx
```diff
@@ -0,0 +1,38 @@
...
+          <EthChainIcon /> Ethereum chain
```
elias-garcia, line 24: If we are going to return an icon with text, I would use proper HTML elements:

```
return (
  <div>
    <EthChainIcon />
    <Typography type="body">Ethereum chain</Typography>
  </div>
)

File: src/views/transaction-details/transaction-details.view.tsx [hunk 3]
```diff
@@ -0,0 +1,117 @@
...
+import Header from "src/views/shared/header/header.view";
+import { ReactComponent as NewWindowIcon } from "src/assets/icons/new-window.svg";
+import { ReactComponent as SpinnerIcon } from "src/assets/icons/spinner.svg";
+import Typography from "../shared/typography/typography.view";
+import { StatusText } from "../activity/components/transaction-card/transaction-card.view";
+import { demoData } from "../activity/demo-data";
+import { getTimeFromNow } from "src/utils/time";
+import { convertTokenAmountToFiat } from "src/utils/amounts";
+import TokenIcon from "../shared/token-icon/token-icon";
+import ChainRow from "./components/chain-row/chain-row";
```
elias-garcia, line 15: Most of the imports are relative, they should be absolute

src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const TransactionDetails: FC = () => {
+  const { transactionId } = useParams();
+  const data = demoData.find((demo) => demo.id.toString() === transactionId);
+  const classes = useTransactionDetailsStyles({ status: data?.status || "initiated" });
+
+  if (!data) {
+    Navigate({ to: "/activity", replace: true });
```
elias-garcia, line 23: Instead of this and returning null, why not:

```ts
return <Navigate to="/activity" options={{replace: true}} />
```
```

File: src/views/activity/components/transaction-card/transaction-card.view.tsx [hunk 5]
```diff
@@ -4,18 +4,20 @@ import useTransactionCardStyles from "src/views/activity/components/transaction-
...
 import Typography from "src/views/shared/typography/typography.view";
 import Card from "src/views/shared/card/card.view";
 import { getTimeFromNow, TimeFromNowParams } from "src/utils/time";
 import { convertTokenAmountToFiat } from "src/utils/amounts";
+import { useNavigate } from "react-router-dom";
+import routes from "src/routes";
+import TokenIcon from "src/views/shared/token-icon/token-icon";
 
-enum StatusText {
+export enum StatusText {
```
elias-garcia, line 15: As now this enum is shared I would write it in the `domain` module, next to the `WalletName`. Also, remember that enum properties should be in MAYUS and with underscores:

```ts
enum StatusText {
  ON_HOLD = "On Hold",
  INITIATED = "Initiated",
  PROCESSING = "Processing",
  FAILED = "Error",
  COMPLETED = "Completed"
}
```

elias-garcia:
Why has the checksums of all the dependencies in the `package-lock` have changed?

src/views/transaction-details/components/chain-row/chain-row.tsx
adrifdez: It's wrapped by `Typography`in `transaction-details.view`

src/views/activity/components/transaction-card/transaction-card.view.tsx
adrifdez: This is waiting to receive data from API so properties maybe can't be in uppercase

adrifdez:
> Why the checksums of all the dependencies in the `package-lock` have changed?

I'm having problems with the downgrade of `react script` and that one was a try to solve it

src/views/activity/components/transaction-card/transaction-card.view.tsx
elias-garcia: But enum properties can not be in lowercase, if they should be like this it should be a data structure but not an enum

src/views/transaction-details/components/chain-row/chain-row.tsx
elias-garcia: Then the wrapping should be done here

src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -0,0 +1,112 @@
...
+import TokenIcon from "src/views/shared/token-icon/token-icon";
+import ChainRow from "src/views/transaction-details/components/chain-row/chain-row";
+
+const TransactionDetails: FC = () => {
+  const { transactionId } = useParams();
+  const data = demoData.find((demo) => demo.id.toString() === transactionId);
+  const classes = useTransactionDetailsStyles({ status: data?.status || "initiated" });
+
+  if (!data) {
+    return Navigate({ to: "/activity", replace: true });
```
elias-garcia, line 23: `Navigate` is a React component, it should be instantiated properly as a React component

src/views/activity/components/transaction-card/transaction-card.view.tsx
adrifdez: I've changed it to object in that case

src/views/transaction-details/components/chain-row/chain-row.tsx
amonsosanz: An icon wrapped by typography?

src/views/transaction-details/transaction-details.styles.ts
```diff
@@ -0,0 +1,116 @@
+import { createUseStyles } from "react-jss";
+
+import { Theme } from "src/styles/theme";
+import { TransactionCardProps } from "src/views/activity/components/transaction-card/transaction-card.view";
+
+interface StylesProps {
+  status: TransactionCardProps["status"];
```
amonsosanz, line 7: We should not couple these two components for this reason.

src/views/transaction-details/transaction-details.styles.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+const useTransactionDetailsStyles = createUseStyles((theme: Theme) => ({
+  card: {
+    padding: theme.spacing(3),
+  },
+  dot: {
+    width: theme.spacing(0.75),
+    height: theme.spacing(0.75),
```
amonsosanz, line 16: Using theme spacing to define sizes will produce unexpected size changes when one decides to change the spacing.

src/views/transaction-details/transaction-details.styles.ts
```diff
@@ -0,0 +1,116 @@
...
+    alignItems: "center",
+    flexDirection: "column",
+    gap: theme.spacing(1),
+    borderBottom: `1px solid ${theme.palette.grey.light}`,
+    paddingBottom: theme.spacing(3),
+    marginBottom: theme.spacing(1.5),
+  },
+  tokenIcon: {
+    width: theme.spacing(6),
+    height: theme.spacing(6),
```
amonsosanz, line 42: Using theme spacing to define sizes will produce unexpected size changes when one decides to change the spacing.

src/views/transaction-details/transaction-details.styles.ts
```diff
@@ -0,0 +1,116 @@
...
+    [theme.breakpoints.upSm]: {
+      margin: [theme.spacing(6), 0],
+    },
+  },
+  finaliseButton: {
+    display: "inline-flex",
+    justifyContent: "center",
+    alignItems: "center",
+    padding: [theme.spacing(2), theme.spacing(2)],
+    width: theme.spacing(31),
```
amonsosanz, line 92: Using theme spacing to define sizes will produce unexpected size changes when one decides to change the spacing.

src/views/transaction-details/components/chain-row/chain-row.style.ts
```diff
@@ -0,0 +1,12 @@
+import { createUseStyles } from "react-jss";
+
+import { Theme } from "src/styles/theme";
+
+const useChainRowStyles = createUseStyles((theme: Theme) => ({
+  hermezChain: {
+    width: theme.spacing(2.5),
```
amonsosanz, line 7: Using theme spacing to define sizes will produce unexpected size changes when one decides to change the spacing.

src/views/transaction-details/components/chain-row/chain-row.tsx
```diff
@@ -0,0 +1,40 @@
...
+        </Typography>
+      );
+    case "hermez":
+      return (
+        <Typography type="body1" className={className}>
+          <HermezChainIcon className={classes.hermezChain} /> Polygon Hermez chain
+        </Typography>
+      );
+    default:
+      return null;
```
amonsosanz, line 36: This feels like a [smell](https://martinfowler.com/bliki/CodeSmell.html). We should not need this default.

src/views/transaction-details/transaction-details.styles.ts
```diff
@@ -0,0 +1,116 @@
...
+    border: "none",
+    backgroundColor: theme.palette.primary.main,
+    cursor: "pointer",
+    borderRadius: theme.spacing(5),
```
amonsosanz, line 96: Using theme spacing to define sizes will produce unexpected size changes when one decides to change the spacing.

src/views/transaction-details/transaction-details.styles.ts
```diff
@@ -0,0 +1,116 @@
...
+      backgroundColor: theme.palette.disabled,
+      cursor: "initial",
+    },
+  },
+  finaliseButtonText: {
+    color: theme.palette.white,
+  },
+  finaliseSpinner: {
+    "& path": {
+      fill: theme.palette.white,
```
amonsosanz, line 111: I can't tell I'm against this, but it's definitely not a common pattern. We normally would have created the icon `spinner-white.svg`. In this case though, since the icon only has a path, I guess it's ok.

src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <Typography type="body2" className={classes.alignRow}>
+            {status === "initiated" ? "Confirmation expected in" : "Date"}
+          </Typography>
+          <Typography type="body1" className={classes.alignRow}>
+            {status === "initiated" ? (
+              <>
+                Estimating Time <SpinnerIcon />
+              </>
+            ) : (
+              <>{getTimeFromNow({ timestamp })} ago</>
```
amonsosanz, line 55: I find the time ago very useful, but I think we must provide a way to see the full date.

I'm of the opinion that it is best to put them side by side, but at least, I would show the actual date in a tooltip.

amonsosanz:
image

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -4,19 +4,21 @@ import useTransactionCardStyles from "src/views/activity/components/transaction-
...
+import routes from "src/routes";
+import TokenIcon from "src/views/shared/token-icon/token-icon";
 
-enum StatusText {
-  "on-hold" = "On Hold",
-  initiated = "Initiated",
-  processing = "Processing",
-  failed = "Error",
-}
+export const statusText = {
```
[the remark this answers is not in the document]
adrifdez: Because enum must be in uppercase

src/routes.ts
```diff
@@ -24,6 +25,11 @@ const routes = {
     Component: Activity,
     isPrivate: true,
   },
+  transactionDetails: {
+    path: "/transaction-details/:transactionId",
```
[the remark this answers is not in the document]
adrifdez: I've used the title of the page to create the url

src/views/shared/card/card.view.tsx
```diff
@@ -4,12 +4,17 @@ import useCardStyles from "src/views/shared/card/card.styles";
...
+  onClick?: () => void;
 }
 
-const Card: FC<CardProps> = ({ className, children }) => {
+const Card: FC<CardProps> = ({ className, children, onClick }) => {
   const classes = useCardStyles();
 
-  return <div className={`${classes.card} ${className || ""}`}>{children}</div>;
+  return (
+    <div className={`${classes.card} ${className || ""}`} onClick={onClick}>
```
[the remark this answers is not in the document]
adrifdez: The hover is affecting the whole card, so I understand that we want that behaviour

src/views/transaction-details/components/chain-row/chain-row.tsx
adrifdez: Typescript doesn't like it 
`Type 'Element | undefined' is not assignable to type 'ReactElement<any, any> | null'.`

src/views/activity/components/transaction-card/transaction-card.view.tsx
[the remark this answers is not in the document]
amonsosanz: I don't get what you mean. Can you reformulate?

src/routes.ts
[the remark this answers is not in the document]
amonsosanz: Normal.

We need to think about routing in a separate issue.

src/views/shared/token-icon/token-icon.tsx
```diff
@@ -0,0 +1,19 @@
...
+  token: "eth" | "dai";
+  className?: string;
+}
+
+const tokenIcons = {
+  eth: EthToken,
+  dai: DaiToken,
+};
+const TokenIcon: FC<TokenIconProps> = ({ token, className }) => {
+  const Icon = tokenIcons[token] || tokenIcons["eth"];
```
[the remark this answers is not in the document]
amonsosanz: But why do we need a fallback?

src/views/shared/card/card.view.tsx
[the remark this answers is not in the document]
amonsosanz: Yeah, this reasoning makes sense.

src/views/transaction-details/components/chain-row/chain-row.tsx
amonsosanz: That is the point. Typescript should like what we do. The fact that we need this default value to make Typescript shut up indicates a flaw in type design. We need to change the design.

amonsosanz: When you create this demo data:

```typescript
const demoData = [{
    id: Math.random(),
    type: "l1",
    token: "dai",
    timestamp: Date.now() - Math.floor(Math.random() * 10000),
    status: "on-hold",
    amount: Math.floor(Math.random() * 100),
}]
```
I interpret `type` as the target chain. Is this correct?. This `type` name prop is ambiguous. I'd change it to `target` in our domain to be more clear.

With a clearer naming, we no longer need to pass both `type` and `side` to this component. If I understand the current logic correctly:

```bash
When ("to" & "l1") | ("from" & "l2") we want to render Ethereum
When ("to" & "l2") | ("from" & "l1") we want to render Polygon
```

Why not have a dummy component like this:

```typescript
interface ChainRowProps {
  targetChain: "l1" | "l2";
  className?: string;
}

const ChainRow: FC<ChainRowProps> = ({ targetChain: targetChain, className }) => {
  const classes = useChainRowStyles();

  const EthereumChain = (
    <Typography type="body1" className={className}>
      <EthChainIcon /> Ethereum chain
    </Typography>
  );

  const PolygonHermezChain = (
    <Typography type="body1" className={className}>
      <HermezChainIcon className={classes.hermezChain} /> Polygon Hermez chain
    </Typography>
  );

  return targetChain === "l1" ? EthereumChain : PolygonHermezChain;
};
```

And move the decision making to the instantiator:

```typescript
type TargetChain = "l1" | "l2";

// this assumes `type` as `target`
const { amount, token, timestamp, type, status } = data;
const toTargetChain: TargetChain = type;
const fromTargetChain: TargetChain = type === "l2" ? "l1" : "l2";

....


<div className={classes.row}>
  <Typography type="body2" className={classes.alignRow}>
    From
  </Typography>
  <ChainRow targetChain={fromTargetChain} className={classes.alignRow} />
</div>
<div className={classes.row}>
  <Typography type="body2" className={classes.alignRow}>
    To
  </Typography>
  <ChainRow targetChain={toTargetChain} className={classes.alignRow} />
</div>
```

src/views/transaction-details/components/chain-row/chain-row.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Typography from "src/views/shared/typography/typography.view";
+import useChainRowStyles from "src/views/transaction-details/components/chain-row/chain-row.style";
+
+interface ChainRowProps {
+  type: "l1" | "l2";
+  side: "to" | "from";
+  className?: string;
+}
+
+const ChainRow: FC<ChainRowProps> = ({ type, side, className }) => {
```
amonsosanz, line 14: I'd change the name of this component to just `Chain`.

It's used in a row by `transaction-details.view`, but that has nothing to do with the component.

amonsosanz:
We have a bug in the icons as you can see here:

image

Details on the issue [here](https://youtu.be/IbhHCsydN5I).

src/views/transaction-details/transaction-details.styles.ts
elias-garcia: I agree with this

src/views/transaction-details/transaction-details.styles.ts
elias-garcia: I think that It's better to do this than having an icon for each color that we need

src/views/transaction-details/components/chain-row/chain-row.tsx
elias-garcia: I agree with this as well

src/views/transaction-details/transaction-details.view.tsx
elias-garcia: We can't decide here, it's a design decission. By the moment I wouldn't worry about this

src/views/activity/components/transaction-card/transaction-card.view.tsx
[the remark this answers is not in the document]
elias-garcia: He needs the keys in lowercase, and enum keys should be in uppercase and without hyphens. That's why we moved this to a constant

src/views/shared/token-icon/token-icon.tsx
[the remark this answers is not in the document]
elias-garcia: It's not relevant by the moment. We will need a fallback but this library is probably going to dissapear

src/views/activity/components/transaction-card/transaction-card.view.tsx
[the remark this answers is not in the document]
amonsosanz: Now I get it! I had to read [this thread](https://github.com/hermeznetwork/bridge-ui/pull/10#discussion_r829840403) to understand what you were talking about.

> As now this enum is shared I would write it in the domain module, next to the WalletName.

You're right here @elias-garcia, this enum belongs to the domain. And you're also right about the naming convention of the enums. I though disagree with the applied solution.

If we need to map a `Status` type with a `StatusText` enum, instead of converting the `enum` to an `object` to be able to use bracket notation, we create a function to do the job:

```typescript
type Status = "processing" | "initiated" | "on-hold" | "completed" | "failed";

export enum StatusText {
  ON_HOLD = "On Hold",
  INITIATED = "Initiated",
  PROCESSING = "Processing",
  FAILED = "Error",
  COMPLETED = "Completed",
}

export function getStatusText(status: Status): StatusText {
  switch (status) {
    case "completed":
      return StatusText.COMPLETED;
    case "failed":
      return StatusText.FAILED;
    case "initiated":
      return StatusText.INITIATED;
    case "on-hold":
      return StatusText.ON_HOLD;
    case "processing":
      return StatusText.PROCESSING;
  }
}
```

src/views/transaction-details/components/chain-row/chain-row.style.ts
```diff
@@ -0,0 +1,12 @@
+import { createUseStyles } from "react-jss";
```
amonsosanz, line 1: This file should be named plural.

src/views/activity/components/transaction-card/transaction-card.view.tsx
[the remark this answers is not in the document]
amonsosanz: Fixed.

src/views/shared/token-icon/token-icon.tsx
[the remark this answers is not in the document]
amonsosanz: Fixed.

src/views/transaction-details/components/chain-row/chain-row.tsx
amonsosanz: Fixed.

src/views/transaction-details/components/chain-row/chain-row.tsx
amonsosanz: Fixed.

amonsosanz:
> We have a bug in the icons as you can see here:
> 
> image
> 
> Details on the issue [here](https://youtu.be/IbhHCsydN5I).

Fixed.

src/views/transaction-details/components/chain-row/chain-row.style.ts
amonsosanz: Fixed.

src/views/transaction-details/transaction-details.styles.ts
amonsosanz: Fixed.

src/views/transaction-details/transaction-details.styles.ts
amonsosanz: Fixed

src/views/transaction-details/transaction-details.view.tsx
amonsosanz: Implemented as requested.

src/domain/index.ts
```diff
@@ -9,6 +9,31 @@ export enum EthereumEvent {
...
+
+export enum TransactionStatusText {
+  PROCESSING = "Processing",
+  INITIATED = "Initiated",
+  ON_HOLD = "On Hold",
+  COMPLETED = "Completed",
+  FAILED = "Error",
+}
+
+export function transactionStatusText(status: TransactionStatus): TransactionStatusText {
```
elias-garcia, line 22: Now that I see this again, I think that it would make more sense to make it a React component. We can delete the enum as well if we move it to a component.

src/views/transaction-details/components/chain/chain.tsx
```diff
@@ -0,0 +1,31 @@
...
+    </Typography>
+  );
+
+  const PolygonHermezChain = (
+    <Typography type="body1" className={className}>
+      <HermezChainIcon className={classes.hermezChain} /> Polygon Hermez chain
+    </Typography>
+  );
+
+  return chain === "ethereum" ? EthereumChain : PolygonHermezChain;
```
elias-garcia, line 28: Why not:


```ts
if (chain === "ethereum") {
  return (
    <Typography type="body1" className={className}>
      <EthChainIcon /> Ethereum chain
    </Typography>
  ) else {
    return (
      <Typography type="body1" className={className}>
        <HermezChainIcon className={classes.hermezChain} /> Polygon Hermez chain
      </Typography>
    )
  }
}
```

src/views/transaction-details/transaction-details.styles.ts
```diff
@@ -0,0 +1,115 @@
+import { createUseStyles } from "react-jss";
+
+import { Theme } from "src/styles/theme";
+import { TransactionStatus } from "src/domain";
+interface StylesProps {
```
elias-garcia, line 5: Missing new line here

src/domain/index.ts
amonsosanz: Enum removed. Only function left.

src/views/transaction-details/components/chain/chain.tsx
amonsosanz: Do you see it simpler? Ok...

amonsosanz: Done

src/views/transaction-details/transaction-details.styles.ts
amonsosanz: Added

elias-garcia (approved):

amonsosanz (approved):

PR #14, opened 2022-03-22
Add infura and jsonrpc providers to get L1 and L2 gas price

amonsosanz:
This PR adds a couple of providers to the global context, one using infura (L1) and one using jsonrpc (L2) so we can use them to get the gas price for L1 and L2.


src/contexts/providers.context.tsx
```diff
@@ -127,35 +145,59 @@ const EthereumProviderProvider: FC = (props) => {
...
+      internalConnectedProvider &&
+      "on" in internalConnectedProvider &&
+      typeof internalConnectedProvider.on === "function"
+    ) {
+      internalConnectedProvider.on(EthereumEvent.ACCOUNTS_CHANGED, onAccountsChanged);
+      internalConnectedProvider.on(EthereumEvent.CHAIN_CHANGED, onChainChangedOrDisconnect);
+      internalConnectedProvider.on(EthereumEvent.DISCONNECT, onChainChangedOrDisconnect);
+    }
+
+    if (env) {
```
elias-garcia, line 158: I would move this block to a new `useEffect` hook on top of this one. I think that it would make it easier to read and each hook will do just one thing:

- One will set the readonly providers.
- The other one will attach and detach provider handlers.

elias-garcia:
Good job Alex! Looks good 💪🏻

src/contexts/providers.context.tsx
amonsosanz: Makes sense. Done. I moved it below though, to avoid having the constants `connectProvider` and `disconnectProvider` far from where they are used.

elias-garcia (approved):

PR #15, opened 2022-03-23
Persist user selected currency

amonsosanz:
This PR implements an adapter to retrieve and store data from/to the local storage and implements the user-selected currency use case on the settings view.

### DIFF
--- a/src/adapters/local-storage.ts
+++ b/src/adapters/local-storage.ts
@@ -15,14 +15,6 @@ export function setCurrency(currency: Currency): void {
   setStorageByKey({ key: constants.PREFERRED_CURRENCY_KEY, value: currency });
 }
 
-export function getStorageVersion(): number {
-  return getStorageByKey({
-    key: constants.STORAGE_VERSION_KEY,
-    defaultValue: constants.STORAGE_VERSION,
-    parser: z.number(),
-  });
-}
-
 // Helpers
 export function getStorageByKey<T>({
   key,
--- a/src/adapters/local-storage.ts
+++ b/src/adapters/local-storage.ts
@@ -1,17 +1,17 @@
 import { z } from "zod";
 
 import * as constants from "src/constants";
-import { Currencies } from "src/domain";
+import { Currency } from "src/domain";
 
-export function getCurrency(): Currencies {
+export function getCurrency(): Currency {
   return getStorageByKey(
     constants.PREFERRED_CURRENCY_KEY,
     constants.PREFERRED_CURRENCY,
-    z.nativeEnum(Currencies)
+    z.nativeEnum(Currency)
   );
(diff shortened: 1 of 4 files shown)


src/adapters/local-storage.ts
```diff
@@ -0,0 +1,49 @@
...
+    defaultValue: constants.PREFERRED_CURRENCY,
+    parser: z.nativeEnum(Currency),
+  });
+}
+
+export function setCurrency(currency: Currency): void {
+  setStorageByKey({ key: constants.PREFERRED_CURRENCY_KEY, value: currency });
+}
+
+export function getStorageVersion(): number {
```
elias-garcia, line 18: I would remove this storage versioning code as we are going to depend on `localStorage` only for this `preferredCurrency` stuff

src/views/settings/settings.view.tsx
```diff
@@ -14,30 +14,30 @@ import { ReactComponent as JpyIcon } from "src/assets/icons/currencies/jpy.svg";
...
+    { icon: <GbpIcon />, name: "GBP", id: Currency.GBP },
+    { icon: <CnyIcon />, name: "CNY", id: Currency.CNY },
+    { icon: <JpyIcon />, name: "JPY", id: Currency.JPY },
   ];
 
   const onDisconnectProvider = () => {
     void disconnectProvider();
   };
 
+  const onCurrencySelected = (currency: Currency) => {
```
elias-garcia, line 36: Maybe `onSelectCurrency` is a better name?

src/constants.ts
```diff
@@ -1,3 +1,11 @@
+import { Currency } from "src/domain";
+
+export const STORAGE_VERSION_KEY = "version";
+export const STORAGE_VERSION = 1;
+
+export const PREFERRED_CURRENCY_KEY = "currency";
+export const PREFERRED_CURRENCY = Currency.USD;
```
elias-garcia, line 7: Maybe `PREFERRED_CURRENCY_DEFAULT`, `DEFAULT_PREFERRED_CURRENCY` or something similar? Just to emphasize that it's the default value.

src/adapters/local-storage.ts
amonsosanz: I was refining some things in the PR and this was one of them.

src/views/settings/settings.view.tsx
amonsosanz: Done

src/constants.ts
amonsosanz: I have consciously omitted the prefix `DEFAULT` since it's implicit in the context of the constants. I mean, do we really need this emphasis?

elias-garcia (approved):

PR #17, opened 2022-03-24
Add integration with a fiat exchange rates API

elias-garcia:
### What does this PR does?

This PR adds the integration with the https://exchangeratesapi.io service, so we are able to convert a token price from USD to the user preferred currency.

### How to test?

To test this feature you can write this code in the `home.view.tsx` file and change the currency in the `Settings` view of the UI. After going back to the home, the value logged in the console should change accordingly.

```ts
const { preferredCurrency, getTokenPrice } = usePriceOracleContext();

useEffect(() => {
  if (preferredCurrency) {
    getTokenPrice("0x7D1AfA7B718fb893dB30A3aBc0Cfc608AaCfeBB0", preferredCurrency)
      .then((tokenPrice) => {
        console.log(tokenPrice);
      })
      .catch((err) => console.log(err));
  }
}, [preferredCurrency, getTokenPrice]);
```

## Checklist

review them:


src/adapters/fiat-exchange-rates.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  return fetch(requestUrl).then(async (res) => {
+    if (res.ok) {
+      const data: unknown = await res.json();
+      const parsedResponse = fiatExchangeRatesParser.safeParse(data);
+
+      if (parsedResponse.success) {
+        return parsedResponse.data.rates;
+      } else {
+        throw new Error("An error occurred when trying to parse the getFiatExchangeRates response");
```
amonsosanz, line 38: We should not loos track of the `parsedResponse.error`. Instead of returning this text, we should return the error, to be later decoded by any interested consumer of this API.

src/adapters/fiat-exchange-rates.ts
```diff
@@ -0,0 +1,53 @@
...
+  };
+}
+
+interface GetFiatExchangeRatesParams {
+  apiKey: string;
+}
+
+const getFiatExchangeRates = ({
+  apiKey,
+}: GetFiatExchangeRatesParams): Promise<FiatExchangeRates> => {
```
amonsosanz, line 22: What about that discussion we had regarding the return of typed errors `Promise<Either<FiatExchangeRates, GetFiatExchangeRatesError>>`?

src/adapters/fiat-exchange-rates.ts
```diff
@@ -0,0 +1,53 @@
...
+        throw new Error("An error occurred when trying to parse the getFiatExchangeRates response");
+      }
+    } else {
+      const error: unknown = await res.json();
+      const parsedResponse = fiatExchangeRatesErrorParser.safeParse(error);
+
+      if (parsedResponse.success) {
+        throw parsedResponse.data.error;
+      } else {
+        throw new Error("An error occured when trying to parse the getFiatExchangeRates error");
```
amonsosanz, line 47: Same as before, I'd better `throw parsedResponse.error`

src/contexts/price-oracle.context.tsx
```diff
@@ -1,29 +1,47 @@
...
 import { useProvidersContext } from "src/contexts/providers.context";
+import { Currency, FiatExchangeRates } from "src/domain";
 import { PriceOracle, PriceOracle__factory } from "src/types/contracts/price-oracle";
+import { getFiatExchangeRates } from "src/adapters/fiat-exchange-rates";
+import { USDT_DECIMALS } from "src/constants";
+import * as storage from "src/adapters/local-storage";
 
 interface PriceOracleContext {
-  getTokenPrice: (tokenAddress: string) => Promise<BigNumber>;
+  preferredCurrency?: Currency;
```
amonsosanz, line 14: I'd make this required and default to `USD`.

src/contexts/price-oracle.context.tsx
```diff
@@ -46,7 +76,25 @@ const PriceOracleProvider: FC = (props) => {
...
-  return <priceOracleContext.Provider value={{ getTokenPrice }} {...props} />;
+  useEffect(() => {
+    if (env) {
+      getFiatExchangeRates({ apiKey: env.REACT_APP_FIAT_EXCHANGE_RATES_API_KEY })
+        .then(setFiatExchangeRates)
+        .catch((error) => {
+          console.error(error);
+          // void parseError(error).then((parsedError) =>
+          //   openSnackbar({ type: "error", parsed: parsedError })
+          // );
```
amonsosanz, line 87: As agreed, we should not open the snackbar from the context so all this should be removed.

src/domain/index.ts
```diff
@@ -39,13 +40,15 @@ export function getTransactionStatusText(status: TransactionStatus): string {
...
-  GBP = "gbp",
-  CNY = "cny",
+  EUR = "EUR",
+  USD = "USD",
+  JPY = "JPY",
+  GBP = "GBP",
+  CNY = "CNY",
 }
 
+export type FiatExchangeRates = Record<string, number>;
```
amonsosanz, line 50: I'd declare this as `Record<keyof typeof Currency, number>` so the resulting type is:

```typescript
type FiatExchangeRates = {
  EUR: number;
  USD: number;
  JPY: number;
  GBP: number;
  CNY: number;
}
```

src/contexts/price-oracle.context.tsx
```diff
@@ -32,7 +50,19 @@ const PriceOracleProvider: FC = (props) => {
...
+      throw new Error("Fiat exchange rates are not available");
+    }
+
+    const rate = await priceOracleContract.getRate(
+      tokenAddress,
+      env.REACT_APP_USDT_CONTRACT_ADDRESS,
+      true
+    );
+    const usdPrice = Number(formatUnits(rate, USDT_DECIMALS));
+    const fiatExchangeRate = fiatExchangeRates[preferredCurrency];
```
amonsosanz, line 63: This could be undefined

src/contexts/price-oracle.context.tsx
```diff
@@ -1,29 +1,47 @@
...
 const priceOracleContextNotReadyErrorMsg = "The price oracle context is not yet ready";
 
 const priceOracleContext = createContext<PriceOracleContext>({
+  changePreferredCurrency: () => {
+    console.error(priceOracleContextNotReadyErrorMsg);
+  },
   getTokenPrice: () => {
     console.error(priceOracleContextNotReadyErrorMsg);
-    return Promise.resolve(BigNumber.from(0));
+    return Promise.resolve(0);
```
amonsosanz, line 27: We should improve the pattern

```typescript
const fooNotReadyErrorMsg = "Foo context is not yet ready";
```
Instead of logging an error and returning a `Promise.resolve()`, in some cases with an arbitrary value, we should:

```typescript
return Promise.reject(fooNotReadyErrorMsg);
```

At the end of the day, if the context is not yet ready we should not resolve the promise.

If you agree, feel free to update this in all the contexts. I can also do it in my current task.

src/contexts/price-oracle.context.tsx
elias-garcia: I don't like it, because the source of the default value should be the `localStorage`, and it's already in charge of setting the `defaultValue` to `Currency.USD`. The point is that the `defaultValue` should be synced with `localStorage` always.

src/domain/index.ts
elias-garcia: Done!

src/contexts/price-oracle.context.tsx
elias-garcia: Done! I prefer the `Partial<Record<keyof typeof Currency, number>>` approach

src/contexts/price-oracle.context.tsx
elias-garcia: Right, done!

src/adapters/parsers.ts
```diff
@@ -2,6 +2,10 @@ import { z } from "zod";
 
 import { StrictSchema } from "src/utils/type-safety";
 import { Env, RouterState } from "src/domain";
+import {
+  GetFiatExchangeRatesError,
+  GetFiatExchangeRatesResponse,
+} from "src/adapters/fiat-exchange-rates";
```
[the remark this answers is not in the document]
elias-garcia: I don't understand what you mean by `cross-dependency coupling`. I was using this file to declare all the parsers.

src/adapters/fiat-exchange-rates.ts
elias-garcia: Right! Fixed

src/adapters/fiat-exchange-rates.ts
elias-garcia: Right! Fixed

src/contexts/price-oracle.context.tsx
amonsosanz: I understand and you're right. I just try to minimize the optional values that this react context architecture is forcing us to adopt.

And at the end of the day, we have a constant `PREFERRED_CURRENCY` so I see not much conflict on using this value as default also here.

src/contexts/price-oracle.context.tsx
elias-garcia: As we talked later, I added the snackbar again until we find a better solution

src/adapters/parsers.ts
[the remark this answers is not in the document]
amonsosanz: Sorry, I just made up this `cross-dependency coupling` term and I expect you to follow...

What I mean is:

We're declaring a couple of parsers in `parsers.ts` that are only relevant to (will only be imported by) `fiat-exchange-rates.ts`, but to create them, we need to import the types `GetFiatExchangeRatesError` and `GetFiatExchangeRatesResponse` from `fiat-exchange-rates.ts` (we need them for the `StrictSchema`).

What I mean by `cross-dependency coupling` is that we are coupling these 2 modules in a circular manner (not hierarchical), since both of them import from the other. This is a [code smell](https://martinfowler.com/bliki/CodeSmell.html) that makes me think that:

> We should consider declaring in `parsers.ts` **only the parsers that are reused by more than one module** of the application and move to the consumer module (in this case `fiat-exchange-rates.ts`) those parsers that only affect that module.

In other words, I'd move the declaration of the parsers `fiatExchangeRatesErrorParser` and ` fiatExchangeRatesParser` to `fiat-exchange-rates.ts`.

src/contexts/price-oracle.context.tsx
amonsosanz: This works for me!

One thing though, are we ok showing this to the user?

image

elias-garcia: It's not very important, take into account that we are going to catch this errors most of the time when testing the ui after deploying it... It's going to pop up just by opening the app. I don't see necessary catching this error and override the message.

src/adapters/parsers.ts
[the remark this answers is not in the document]
elias-garcia: Ah right, I agree! I was using this as a container for all the parsers by the moment. I think that this is outside of the scope of this PR, but I'll move all of them to its module if you want.

src/contexts/price-oracle.context.tsx
elias-garcia: Well, at the end it's okay to be `undefined` at the beginning, until we read it from `localStorage` and we take a decision about its current value. I can add the `PREFERRED_CURRENCY` but it's just a way to trick how Context works...

src/adapters/parsers.ts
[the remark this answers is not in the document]
elias-garcia: Done

adrifdez (approved):

src/adapters/fiat-exchange-rates.ts
amonsosanz: We agreed to give this a try on a different branch.

PR #18, opened 2022-03-25
Downgrade dependency to fix react-script problem

adrifdez:
### What does this PR does?

Downgrade `react-error-overlay` to solve problem with `react-script`

## Checklist

review them:


amonsosanz (approved):
Nice! The lovely overlay is back and the error is gone!

image

elias-garcia (approved):

PR #19, opened 2022-03-25
Integrate bridge api

amonsosanz:
This PR adds a script to build the typescript types and clients from the proto definition of the Bridge API.

To test it, go for example to `Home` and import the `GetBridgesRequest ` and the `BridgeServiceClient` and instantiate both a `request` and a `client`:

```typescript
import { GetBridgesRequest } from "src/types/proto/bridge/v1/query_pb";
import { BridgeServiceClient } from "src/types/proto/bridge/v1/QueryServiceClientPb";

useEffect(() => {
  const API_URL = "https://localhost/";
  const ETH_ADDRESS = "0x99B2d88d2CdA02e70f350A4591dCcC5D52C8C8e3";

  const client = new BridgeServiceClient(API_URL);
  const request = new GetBridgesRequest();

  request.setEtherAddr(ETH_ADDRESS);
  client.getBridges(request, {}).then((bridges) => {
    console.log(bridges);
  }).catch(console.error);
});
```


adrifdez (approved):

package.json
```diff
@@ -10,6 +10,8 @@
     "@types/react-dom": "^17.0.11",
     "@walletconnect/ethereum-provider": "^1.7.5",
     "ethers": "^5.5.4",
+    "google-protobuf": "^3.20.0-rc.2",
+    "grpc-web": "^1.3.1",
```
elias-garcia, line 14: For what are `google-protobuf` and `grpc-web` dependencies used? I'm just curious about them as this is new for me

elias-garcia (approved):

package.json
amonsosanz: For runtime. They are used by the generated JS.

adrifdez (approved):

PR #21, opened 2022-03-29
Add useCallback to Providers

adrifdez:
### What does this PR does?

Convert callbacks to useCallback Hooks to avoid renders and loop problems 

## Checklist

review them:

### DIFF
--- a/src/contexts/global.context.tsx
+++ b/src/contexts/global.context.tsx
@@ -1,4 +1,4 @@
-import { createContext, FC, useContext, useState } from "react";
+import { createContext, FC, useCallback, useContext, useMemo, useState } from "react";
 
 import { Message } from "src/domain";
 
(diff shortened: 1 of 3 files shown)


src/contexts/providers.context.tsx
```diff
@@ -124,7 +127,7 @@ const ProvidersProvider: FC = (props) => {
         resolve();
       });
     }
-  };
+  }, [connectedProvider?.provider]);
```
elias-garcia, line 130: This looks strange to me, shouldn't it be just `connectedProvider`?

src/contexts/global.context.tsx
```diff
@@ -36,20 +36,29 @@ const GlobalProvider: FC = (props) => {
...
+  const closeSnackbar = useCallback(
+    (): void =>
+      setSnackbar({
+        status: "closed",
+      }),
+    []
+  );
 
-  return <globalContext.Provider value={{ snackbar, openSnackbar, closeSnackbar }} {...props} />;
+  const getValue = useMemo(
```
amonsosanz, line 56: I'd name this just `value` since it's a value (not a function)

amonsosanz (approved):

elias-garcia (approved):

PR #23, opened 2022-03-30
Improve the script generate-proto-types.sh

amonsosanz:
* Update the target branch of the repo `hermeznetwork/comms-protocol` to be `main`.
* Improve the script to make sure all the dependencies are met and throw an error otherwise.


scripts/generate-proto-types.sh
```diff
@@ -1,16 +1,70 @@
...
 IMPORTS_PATH="$TMP_PATH/third_party"
 PROTO_PATH="proto/bridge/v1"
 PROTO_FILE="query.proto"
 OUTPUT_PATH="src/types"
 REPOSITORY="https://$GH_TOKEN@github.com/hermeznetwork/comms-protocol.git"
-BRANCH="feature/bridge"
+BRANCH="main"
 
 mkdir $TMP_PATH
 mkdir $OUTPUT_PATH
```
adrifdez, line 70: I think that you can remove this line to avoid a warning line

scripts/generate-proto-types.sh
```diff
@@ -1,16 +1,70 @@
...
+  exit 1
+fi
+
+# All dependencies are available, we can move on!
 TMP_PATH="tmp"
 IMPORTS_PATH="$TMP_PATH/third_party"
 PROTO_PATH="proto/bridge/v1"
 PROTO_FILE="query.proto"
 OUTPUT_PATH="src/types"
 REPOSITORY="https://$GH_TOKEN@github.com/hermeznetwork/comms-protocol.git"
```
adrifdez, line 66: Could we do a selector between token or ssh?

elias-garcia (approved):

scripts/generate-proto-types.sh
elias-garcia: This is a good idea btw!

scripts/generate-proto-types.sh
amonsosanz: Right. I removed the creation of both directories since they are automatically created by `git` and `protoc`.

scripts/generate-proto-types.sh
amonsosanz: Interesting point!

It's funny. The reason I'm using HTTP and a token is that initially, I was just downloading the bridge's `query.proto` file with `curl`, but later I realized that it imports `third_party/google/api/annotations.proto` so I decided to move to `git clone` without even thinking about the SSH option.

adrifdez (approved):

PR #27, opened 2022-03-31
Add GET /bridges/{ethereumAddress} API call

### LINKED ISSUE #26: Retrieve the list of bridges from the API
We need to retrieve the list of current bridges from the API. By the moment we are going to use the mock server.

elias-garcia:
Closes #26.

### What does this PR does?

This PR adds the necessary code to retrieve the `bridges` from the API. It also adds `axios` to the `fiat-exchange-rates-api` module.

### How to test?

First of all you'll need to deploy the Bridge API in your machine using Docker: https://github.com/hermeznetwork/hermez-bridge

After that, you can add this lines to the home view and check if the response is properly logged into the console.

```
const { getBridges } = useBridgeContext();

useEffect(() => {
  getBridges({ ethereumAddress: "0xeB17ce701E9D92724AA2ABAdA7E4B28830597Dd9" })
    .then(console.log)
    .catch(console.error);
}, [getBridges]);
```
## Checklist

review them:


src/adapters/bridge-api.ts
```diff
@@ -0,0 +1,74 @@
...
+    amount: z.string(),
+    dest_net: z.number(),
+    dest_addr: z.string(),
+    deposit_cnt: z.string(),
+  })
+);
+
+const getBridgesResponseParser = StrictSchema<GetBridgesResponse>()(
+  z.object({
+    deposits: z.optional(z.array(bridgeParser)),
```
amonsosanz, line 36: In the wallet, we agreed on the policy of not failing to decode a list when a single element of the list does not meet the interface. Instead, we agreed to omit the offender. I see that this is no longer honored here. Is this decision conscious?

src/adapters/bridge-api.ts
```diff
@@ -0,0 +1,74 @@
...
+
+interface Bridge {
+  token_addr: string;
+  amount: string;
+  dest_net: number;
+  dest_addr: string;
+  deposit_cnt: string;
+}
+
+interface GetBridgesParams {
```
amonsosanz, line 15: I'd like to open a discussion regarding these whatever[Params] named interfaces.

Being interfaces to type the params object of some helper and accessory functions, would you agree on declaring them inline @elias-garcia and @adrifdez?

elias-garcia: What's the advantage of declaring them inline?

src/adapters/bridge-api.ts
elias-garcia: Yes, this API is different. This is a readonly API and they read the data directly from the smart contract, so if one element of the list is wrong, most likely the entire list is going to be wrong. So I don't see the need of having unknown parsers, reducers etc to filter wrong elements of the list in this case.

src/adapters/bridge-api.ts
```diff
@@ -0,0 +1,74 @@
...
+  dest_addr,
+  dest_net,
+  deposit_cnt,
+}: Bridge): domain.Bridge => ({
+  tokenAddress: token_addr,
+  amount: amount,
+  destinationAddress: dest_addr,
+  destinationNetwork: dest_net,
+  depositCount: deposit_cnt,
+});
```
amonsosanz, line 52: Instead of using this _Persistence to Domain_ transformation function as an external artifact, I'd rather use [Zod's built in transformation](https://github.com/colinhacks/zod#transform) mechanism:

```typescript
const bridgeParser = StrictSchema<domain.Bridge, Bridge>()(
  z
    .object({
      token_addr: z.string(),
      amount: z.string(),
      dest_net: z.number(),
      dest_addr: z.string(),
      deposit_cnt: z.string(),
    })
    .transform(({ token_addr, amount, dest_addr, dest_net, deposit_cnt }) => ({
      tokenAddress: token_addr,
      amount: amount,
      destinationAddress: dest_addr,
      destinationNetwork: dest_net,
      depositCount: deposit_cnt,
    }))
);
```

Like this we rely on the parser itself to transform the data to our domain, making sure that whenever we use it to decode the API input, we get our domain entity as output.

Note that for this to work we need to improve a bit our `StrictSchema` helper to accept two types, one as output and a new one, which is optional, as input: `StrictSchema<domain.Bridge, Bridge>`:

```typescript
export const StrictSchema: <O, I = O>() => <UO, UI = UO>(
  u: Exact<ZodSchema<O, ZodTypeDef, I>, ZodSchema<UO, ZodTypeDef, UI>> extends true
    ? Exact<Required<O>, Required<UO>> extends true
      ? ZodSchema<UO, ZodTypeDef, UI>
      : never
    : never
) => ZodSchema<O, ZodTypeDef, I> =
  () =>
  <O, I = O>(u: unknown) =>
    u as ZodSchema<O, ZodTypeDef, I>;
```

Like this, we optionally accept an _Input_ type parameter which, as Zod does, defaults to the already required _Output_ type parameter and provide it to the `ZodSchema`.

package.json
```diff
@@ -35,7 +36,6 @@
     "build": "react-scripts build",
     "format": "prettier --write .",
     "generate-contract-types": "./scripts/generate-contract-types.sh",
-    "generate-proto-types": "./scripts/generate-proto-types.sh -ssh",
     "lint": "eslint --fix --ext .ts,.tsx .",
     "postinstall": "npm run generate-contract-types && npm run generate-proto-types",
```
amonsosanz, line 40: `&& npm run generate-proto-types` should be now removed

src/contexts/bridge.context.tsx
```diff
@@ -0,0 +1,44 @@
+import { createContext, FC, useCallback, useContext } from "react";
+
+import { useEnvContext } from "src/contexts/env.context";
+import { Bridge } from "src/domain";
+import * as api from "../adapters/bridge-api";
```
amonsosanz, line 5: Didn't you change your VSCode config to default importing absolute routes?

package.json
elias-garcia: Right! Done

src/contexts/bridge.context.tsx
elias-garcia: I import it manually I think, because VS Code doesn't support this `import * as api`! That's why it was wrong, fixed!

src/adapters/bridge-api.ts
elias-garcia: > - Assure they are contextualized (i.e close to where they are used/required)

I don't see this a pro... At the end, if we do this always we would end up with `React.FC<{ // props }>` inline as well and much more things. The problem is the readability, take a look how the inline looks with more params and one more function:

Captura de Pantalla 2022-04-04 a las 20 16 13

In comparison with declaring the interface on top:

Captura de Pantalla 2022-04-04 a las 20 15 12

Now imagine if you need to add more complex types like generics or using `&`, `|` operators...

> - Get better hints in VSCode

Using separate interfaces you'll have the same hints if you hit `ctrl + space`:

Captura de Pantalla 2022-04-04 a las 20 13 11

From my point of view I prefer to declare types and interfaces separately because they make the code more readable and easier to compose different types.

src/adapters/bridge-api.ts
elias-garcia: The only thing that I don't like of this approach is that is specific of `zod` and it's not portable, while the other is just a `map` and a function (which is portable to any parsing library). Which could be suitable for keeping it reusable is keeping the function and doing this:

```ts
const bridgeParser = StrictSchema<domain.Bridge, Bridge>()(
  z
    .object({
      token_addr: z.string(),
      amount: z.string(),
      dest_net: z.number(),
      dest_addr: z.string(),
      deposit_cnt: z.string(),
    })
    .transform(transformApiBridgeToDomain)
);
```

Are you okay with this approach?

amonsosanz: Of course! The important thing is that when you use the parser you get the domain entity as output.

I'd though remove the prefix `transform` from the function name, so it reads `.transform(apiBridgeToDomain)`.

The name `apiBridgeToDomain` already implies a transformation.

src/adapters/bridge-api.ts
amonsosanz: This reads as if you were not using an object at all:



You see the name of the function and then, one param per line. The only difference is the `params` key which we could agree to call just `p` or `_`. It's by far more readable than this:



Like this, I have to go to the interface `GetBridgesParams` to find the actual params wrapped in the object. Yes, I can use control and mitigate the problem for small, not deeply nested interfaces, but why having the problem at all?

The drawback of passing one object to the functions is precisely that we lose context and readability since we can no longer read the params in the context we need them: the signature of the function.

.env.example
```diff
@@ -1,6 +1,7 @@
 REACT_APP_INFURA_API_KEY=
 REACT_APP_L1_PROVIDER_NETWORK=rinkeby
 REACT_APP_L2_PROVIDER_URL=http://localhost:8545
-REACT_APP_PRICE_ORACLE_CONTRACT_ADDRESS=0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6
+REACT_APP_UNISWAP_QUOTER_CONTRACT_ADDRESS=0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6
 REACT_APP_USDT_CONTRACT_ADDRESS=0xdAC17F958D2ee523a2206206994597C13D831ec7
 REACT_APP_FIAT_EXCHANGE_RATES_API_KEY=
+REACT_APP_BRIDGE_API_URL=http://localhost:9090
```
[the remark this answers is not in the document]
elias-garcia: Right, fixed!

src/adapters/bridge-api.ts
```diff
@@ -0,0 +1,74 @@
+import axios from "axios";
```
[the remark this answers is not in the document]
amonsosanz: Fixed using the image from `hermeznetwork/hermez-bridge:latest`

amonsosanz (approved):

PR #29, opened 2022-04-01
Integrate bridge contract

amonsosanz:
This PR integrates the contract of the bridge and makes it accessible through a new `bridge` React Context.

To test it you will need to start a local HardHat network and deploy the contracts of the bridge. [Follow these instructions](https://github.com/hermeznetwork/internal-tools/tree/main/FE-tools).

You will also need to set up this network in MetaMask:

image

And add this code to the Home page:

```typescript
import { useEffect } from "react";
import { useBridgeContext } from "src/contexts/bridge.context";
import { BigNumber } from "ethers";
import { ethers } from "ethers";

const { bridge, claim } = useBridgeContext();

useEffect(() => {
  const amount = BigNumber.from("0x0de0b6b3a7640000");
  const bridgeDestinationNetwork = 1;
  const destinationAddress = "0x730F01feBde4caBcBa8240020854610BAC89BAb2";
  const claimOriginalNetwork = "1";
  const claimDestinationNetwork = 0;
  const smtpProof = (() => {
    try {
      return JSON.parse(prompt("smtpProof") || "[]") as Array<string>;
    } catch (error) {
      return [];
    }
  })(); // from API's merkle_proof
  const globalExitRootNum = parseInt(prompt("globalExitRootNum") || "0"); // from API's exit_root_num
  if (isNaN(globalExitRootNum) === false) {
    const mainnetExitRoot = prompt("mainnetExitRoot") || ""; // from API's main_exit_root
    const rollupExitRoot = prompt("rollupExitRoot") || ""; // from API's rollup_exit_root
    void claim(
      ethers.constants.AddressZero,
      amount,
      claimOriginalNetwork,
      claimDestinationNetwork,
      destinationAddress,
      smtpProof,
      globalExitRootNum,
      mainnetExitRoot,
      rollupExitRoot
    )
      .then((claimTx) => {
        console.log("Claim response TX:", claimTx);
        void bridge(
          ethers.constants.AddressZero,
          amount,
          bridgeDestinationNetwork,
          destinationAddress
        )
          .then((bridgeTx) => {
            console.log("Bridge response TX:", bridgeTx);
          })
          .catch(console.error);
      })
      .catch(console.error);
  }
}, [bridge, claim]);
```

After logging in, the UI will prompt for a `smtpProof`, a `globalExitRootNum`, a `mainnetExitRoot` and a `rollupExitRoot`. You can get this data from the file `claim_output.json` that is generated by the command `node generateClaimBridge.js -o 1 -a 1 -d 0x730F01feBde4caBcBa8240020854610BAC89BAb2` of the FE-tools. You can `cat claim_output.json` and copy each value.

When these four parameters are provided, MetaMask will ask you to sign both transactions. The result will be logged in the console.

https://user-images.githubusercontent.com/15896177/161330916-f4e93f47-6b32-43ba-99a2-49f428b23e74.mp4

In case you're getting a "Nonce too high" error, reset your MM account as follows:

nonce-too-high


.env.example
```diff
@@ -1,6 +1,7 @@
 REACT_APP_INFURA_API_KEY=
 REACT_APP_L1_PROVIDER_NETWORK=rinkeby
 REACT_APP_L2_PROVIDER_URL=http://localhost:8545
+REACT_APP_BRIDGE_CONTRACT_ADDRESS=
```
elias-garcia, line 4: We can add the local deployment address to this env var.

src/contexts/bridge.context.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useProvidersContext } from "src/contexts/providers.context";
+import { Bridge, Bridge__factory } from "src/types/contracts/bridge";
+
+interface BridgeContext {
+  bridge: (
+    token: string,
+    amount: BigNumber,
+    destinationNetwork: number,
+    destinationAddress: string,
+    overrides?: PayableOverrides & { from?: string | Promise<string> }
```
elias-garcia, line 14: I wouldn't expose this `overrides` param as we are not going to use it.

src/contexts/bridge.context.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    if (bridgeContract === undefined) {
+      throw new Error("Bridge contract is not available");
+    }
+
+    return bridgeContract.bridge(token, amount, destinationNetwork, destinationAddress, overrides);
+  };
+
+  const claim = async (
```
elias-garcia, line 63: Shouldn't this and `bridge` functions use the `useCallback` hook as they rely on state from other hooks?

src/contexts/bridge.context.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return Promise.reject(bridgeContextNotReadyErrorMsg);
+  },
+});
+
+const BridgeProvider: FC = (props) => {
+  const env = useEnvContext();
+  const { connectedProvider } = useProvidersContext();
+  const [bridgeContract, setBridgeContract] = useState<Bridge>();
+
+  const bridge = (
```
elias-garcia, line 45: Not for this PR, but I think that it would be useful to use object params instead of multiple arguments in the contexts as well, at the end they are like "`utils`" that we are calling from different places.

src/contexts/bridge.context.tsx
amonsosanz: I just enter the PR to comment on this 😄

src/contexts/bridge.context.tsx
amonsosanz: They are being used [as you can see here](https://github.com/hermeznetwork/internal-tools/blob/main/FE-tools/generateClaimBridge.js#L58). And more important, if you miss them the contract returns an error.

src/contexts/bridge.context.tsx
amonsosanz: Sure, agreed. I can change it in this PR, but I will wait till we make a decision on [this comment](https://github.com/hermeznetwork/bridge-ui/pull/27/files#r841504240
).

src/contexts/bridge.context.tsx
elias-garcia: Right, thanks for pointing that! The things it that we only need to send this `{ value: amount }` when we are sending `ether`. If we are sending an ERC-20 token this is not needed. As I proposed in Slack, I'll keep this simple because we are missing the ERC-20 approval as well, so I'll say that we only support `ETH` bridges in this iterarion:

```ts
if (tokenAddress !== ethers.constants.AddressZero) { {
  throw new Error("ERC-20 tokens are not supported yet")
}

return bridgeContract.bridge(tokenAddress, amount, destinationNetwork, destinationAddress, { value: amount });
```

I'm creating another issues to support ERC-20 tokens on other iteration.

src/contexts/bridge.context.tsx
elias-garcia: Nice, going to take a look at it! But I'll do it in another PR to be able to review it easily.

src/contexts/bridge.context.tsx
amonsosanz: Ok, that makes sense. Done!

elias-garcia (approved):

adrifdez (approved):

PR #54, opened 2022-04-26
Estimate bridge fee

elias-garcia:
### What does this PR does?

This PR improves some things:

1. Adds the `estimatedFee` of the `bridge` smart contract call to the `TransactionConfirmation` view (both L1 and L2).
2. Removes the `estimatedFee` for the `claim` temporarily as it's very hard to estimate it without having the `merkleProof` and some other required params for the tx.
3. Adds support to call `bridge` and `claim` functions depending on the chain you are connected to. When we merge #53 we should prompt here the user to switch to the proper network if it's not already connected to it.
4. Adds an error to the `Login` view if you are not connected to one of the supported chains.
5. Adds the `bridgeNetworkId` to the `Chain` domain entity, so we are able to send it in the `bridge` and `claim` functions.

## Checklist

review them:


src/contexts/providers.context.tsx
```diff
@@ -48,23 +58,39 @@ const ProvidersProvider: FC = (props) => {
...
-                    setAccount({ status: "pending" });
-                  } else {
-                    openSnackbar({
-                      type: "error",
-                      parsed: errorMsg,
-                    });
-                  }
+
+            return web3Provider.getNetwork().then((network) => {
+              if (env === undefined) {
```
adrifdez, line 63: Could be this check above this resolve?

src/views/home/components/list/list.view.tsx
```diff
@@ -59,33 +60,44 @@ const List: FC<ListProps> = ({ placeholder, list, onClose }) => {
...
+
   return (
     <Portal>
       <div className={classes.background} onClick={onOutsideClick}>
         <Card className={classes.card}>
           <div className={classes.search}>
             <SearchIcon />
-            <input className={classes.input} placeholder={placeholder} onChange={onChange} />
+            <input
+              ref={inputRef}
```
adrifdez, line 76: We could remove `Search Network input`  while we've just two supported networks

src/constants.ts
```diff
@@ -22,22 +22,30 @@ export const REPORT_ERROR_FORM_URL =
 
 export const UNISWAP_V3_POOL_FEE = 3000;
 
+export const BRIDGE_ETHEREUM_NETWORK_ID = 0;
```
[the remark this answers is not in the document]
elias-garcia: Right! Removed

src/contexts/providers.context.tsx
elias-garcia: Done!

src/views/home/components/list/list.view.tsx
elias-garcia: I agree, it was on my plan. I've commented it temporarily.

adrifdez:
Could we center this error?
Screenshot_20220428_120654

src/views/home/components/list/list.view.tsx
```diff
@@ -28,64 +27,45 @@ interface ListProps {
...
-
-  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const value = event.target.value.trim();
-    const filtered: List =
-      list.type === "chain" ? filterChainList(list, value) : filterTokenList(list, value);
-
-    setValues(filtered);
-  };
+  const [values] = useState(list);
+  const inputRef = useRef<HTMLInputElement>(null);
```
adrifdez, line 33: it's not used after remove SearchBox

elias-garcia:
> Could we center this error? Screenshot_20220428_120654

In Figma it's aligned to the left.

adrifdez:
First gray box has more top-margin than last box bottom-margin
Screenshot_20220428_121024

src/views/home/components/transaction-form/transaction-form.styles.ts
```diff
@@ -48,11 +48,7 @@ const useTransactionFormStyles = createUseStyles((theme: Theme) => ({
     padding: [theme.spacing(1.75), theme.spacing(1)],
     backgroundColor: theme.palette.grey.light,
     borderRadius: 8,
-    // cursor: "pointer",
     border: "none",
-    // "&:hover": {
```
adrifdez, line 53: Better remove it than comment it

src/views/home/components/transaction-form/transaction-form.view.tsx
```diff
@@ -129,15 +146,9 @@ const TransactionForm: FC<TransactionFormProps> = ({ onSubmit }) => {
         <div className={classes.row}>
           <div className={classes.box}>
             <Typography type="body2">To</Typography>
-            <div
-              className={classes.chainSelector}
-              // onClick={() =>
-              //   setList({ type: "chain", items: env.chains, onClick: onChainToButtonClick })
-              // }
```
adrifdez, line 136: We could remove this part as well

src/contexts/providers.context.tsx
```diff
@@ -133,6 +165,27 @@ const ProvidersProvider: FC = (props) => {
...
+      return chain.provider.getFeeData().then((feeData) => {
+        if (feeData.maxFeePerGas !== null) {
+          return gasUnits.mul(feeData.maxFeePerGas);
+        }
+
+        if (feeData.gasPrice !== null) {
+          return gasUnits.mul(feeData.gasPrice);
+        }
+
+        return undefined;
```
amonsosanz, line 183: Should'n we throw instead of returning `undefined`? If I understood well, we must either get `maxFeePerGas` or `gasPrice` so both being `null` looks like an error.

src/views/home/components/list/list.view.tsx
amonsosanz: I guess this was to put the focus back on the SearchBox. How's it that the SearchBox has been removed?

src/views/home/components/transaction-form/transaction-form.view.tsx
```diff
@@ -40,9 +40,11 @@ const TransactionForm: FC<TransactionFormProps> = ({ onSubmit, transaction, acco
   const onChainFromButtonClick = (from: Chain) => {
     if (env && formData) {
       const to = env.chains.find((chain) => chain.key !== from.key);
+
       if (to) {
         setFormData({ ...formData, from, to, amount: undefined });
         setList(undefined);
+        setFormData({ ...formData, from, to });
```
amonsosanz, line 47: Duplicated. Merge issue I guess

src/contexts/bridge.context.tsx
```diff
@@ -26,28 +31,36 @@ interface GetClaimsParams {
   ethereumAddress: string;
 }
 
+interface BridgeParams {
+  chain: Chain;
+  token: Token;
+  amount: BigNumber;
+  destinationChain: Chain;
```
amonsosanz, line 38: I'd rename this as `to` and the above `chain` as `from`

src/domain/index.ts
```diff
@@ -123,4 +124,5 @@ export interface TransactionData {
   to: Chain;
   token: Token;
   amount: BigNumber;
+  estimatedFee?: BigNumber;
```
amonsosanz, line 127: This is never used. How's that?

src/contexts/bridge.context.tsx
```diff
@@ -130,68 +145,108 @@ const BridgeProvider: FC = (props) => {
     [env]
   );
 
+  const estimateBridgeGasPrice = useCallback(
```
amonsosanz, line 148: This function is never used, how's that?

src/views/home/components/list/list.view.tsx
adrifdez: Because you see all the supported chains without use SearchBox.

amonsosanz: Fair enough, but then why did we add a filter? 🤣

adrifdez: Design and the box was reused in Tokenlist

src/domain/index.ts
elias-garcia: This doesn't apply anymore

src/contexts/providers.context.tsx
elias-garcia: Right! Fixed

src/views/home/components/transaction-form/transaction-form.view.tsx
elias-garcia: This is outdated

src/contexts/bridge.context.tsx
```diff
@@ -130,68 +154,114 @@ const BridgeProvider: FC = (props) => {
...
-      if (token === ethersConstants.AddressZero) {
-        return bridgeContract.bridge(token, amount, destinationNetwork, destinationAddress, {
-          value: amount,
+      return contract.estimateGas
+        .bridge(token.address, amount, to.bridgeNetworkId, destinationAddress, {
+          ...overrides,
+          from: destinationAddress,
+        })
+        .then((gasLimit) => {
+          const gasIncrease = gasLimit.div(BRIDGE_CALL_GAS_INCREASE).toNumber();
```
amonsosanz, line 181: Why convert it `toNumber()` if it's later added to another BigNumber?

src/contexts/providers.context.tsx
```diff
@@ -49,35 +61,57 @@ const ProvidersProvider: FC = (props) => {
...
+              );
+              const requestedChainId = requestedNetwork.chainId;
+              const supportedChainIds = supportedNetworks.map((network) => network.chainId);
+
+              if (!supportedChainIds.includes(requestedChainId)) {
+                return setAccount({
+                  status: "failed",
+                  error: "Switch your network to Ethereum or Polygon Hermez to continue",
+                });
+              }
```
amonsosanz, line 86: https://user-images.githubusercontent.com/15896177/165972428-e3c56c69-9794-4f81-8a7b-69270235403f.mp4

src/contexts/providers.context.tsx
```diff
@@ -49,35 +61,57 @@ const ProvidersProvider: FC = (props) => {
...
+            } else {
+              return setAccount({
+                status: "failed",
+                error: `We cannot detect your wallet. Make sure the ${WalletName.METAMASK} extension is installed and active in your browser`,
+              });
+            }
+          } catch (error) {
+            const errorMessage = await parseError(error);
+
+            if (!isMetamaskUserRejectedRequestError(error)) {
```
amonsosanz, line 101: In which case are we going to get a `MetamaskUserRejectedRequestError` considering that we're not asking for any signature in the login process?

src/contexts/providers.context.tsx
```diff
@@ -140,6 +174,43 @@ const ProvidersProvider: FC = (props) => {
     }
   }, [connectedProvider]);
 
+  const estimateGasPrice = useCallback(
+    ({ chain, gasUnits }: EstimateGasPriceParams): Promise<BigNumber> => {
+      if (chain.provider === undefined) {
```
amonsosanz, line 179: `chain.provider` can't be `undefined`

src/contexts/providers.context.tsx
```diff
@@ -140,6 +174,43 @@ const ProvidersProvider: FC = (props) => {
...
+        throw new Error("Provider is not available");
+      }
+
+      return chain.provider.getFeeData().then((feeData) => {
+        if (feeData.maxFeePerGas !== null && feeData.maxPriorityFeePerGas !== null) {
```
amonsosanz, line 184: Why check that `feeData.maxPriorityFeePerGas !== null` if we don't use `maxPriorityFeePerGas`?

src/views/home/components/amount-input/amount-input.view.tsx
```diff
@@ -15,25 +15,28 @@ interface AmountInputProps {
...
   balance: BigNumber;
-  fee: BigNumber;
+  fee?: BigNumber;
   onChange: (params: onChangeParams) => void;
 }
 
 const AmountInput: FC<AmountInputProps> = ({ value, token, balance, fee, onChange }) => {
   const defaultInputValue = value ? formatUnits(value, token.decimals) : "";
   const [inputValue, setInputValue] = useState(defaultInputValue);
+  const [actualFee, setActualFee] = useState<BigNumber>();
```
amonsosanz, line 25: We should initialize this as zero since later, in the effect, when `fee` is `undefined` we set it as zero.

Setting zero as the initial value we get rid of the `undefined` case and avoid having to initialize it asynchronously later in the effect.

src/contexts/providers.context.tsx
```diff
@@ -140,6 +174,43 @@ const ProvidersProvider: FC = (props) => {
...
+  const isConnectedProviderChainOk = useCallback(
+    async (chain: Chain): Promise<boolean> => {
+      if (connectedProvider === undefined) {
+        return false;
+      } else {
+        try {
+          const connectedProviderNetwork = await connectedProvider.getNetwork();
+          const chainProviderNetwork = await chain.provider.getNetwork();
+
+          return connectedProviderNetwork === chainProviderNetwork;
```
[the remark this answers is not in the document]
elias-garcia: Right! Fixed

src/contexts/providers.context.tsx
elias-garcia: It's a leftover for a console.log hehe Fixed!

src/contexts/providers.context.tsx
elias-garcia: Fixed!

src/contexts/providers.context.tsx
elias-garcia: If you have not logged in yet to the app, you need to choose which account you want to connect. In that scenario you can click on the "Cancel" button and get this error.

src/contexts/bridge.context.tsx
```diff
@@ -130,68 +154,114 @@ const BridgeProvider: FC = (props) => {
...
+      destinationNetwork,
+      destinationAddress,
+      index,
+      smtProof,
+      globalExitRootNum,
+      mainnetExitRoot,
+      rollupExitRoot,
+    }: ClaimParams): Promise<ContractTransaction> => {
+      if (env === undefined) {
+        throw new Error("Environment is not available");
```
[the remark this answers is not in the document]
elias-garcia: Yep, and copy paste later... I've moved all to just `Env`.

src/contexts/bridge.context.tsx
elias-garcia: Good catch! I used it for logging and forgot to remove it

src/constants.ts
```diff
@@ -23,6 +23,8 @@ export const REPORT_ERROR_FORM_URL =
 
 export const UNISWAP_V3_POOL_FEE = 3000;
 
+export const BRIDGE_CALL_GAS_INCREASE = 10; // 10%
```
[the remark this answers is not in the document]
elias-garcia: Done!

src/adapters/env.ts
```diff
@@ -61,6 +69,7 @@ const envParser = StrictSchema<Env, domain.Env>()(
       REACT_APP_BRIDGE_API_URL: z.string(),
       REACT_APP_L1_BRIDGE_CONTRACT_ADDRESS: z.string(),
       REACT_APP_L2_BRIDGE_CONTRACT_ADDRESS: z.string(),
+      REACT_APP_BRIDGE_POLYGON_HERMEZ_NETWORK_ID: z.string(),
```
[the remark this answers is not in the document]
elias-garcia: Well, it's the network id used in the `Bridge` smart contract. The API returns this networkId as well. `POLYGON_HERMEZ_NETWORK_ID` looks more related to the protocol itself when this is a peculiarity of the bridge dapp. But if it's clear that this `networkId` is related to the bridge contract, I prefer to change it as well as the name is simpler. Let me know and I'll change it.

src/contexts/providers.context.tsx
```diff
@@ -49,35 +61,57 @@ const ProvidersProvider: FC = (props) => {
...
+              setConnectedProvider(web3Provider);
+              return setAccount({ status: "successful", data: accounts[0] });
+            } else {
+              return setAccount({
+                status: "failed",
+                error: `We cannot detect your wallet. Make sure the ${WalletName.METAMASK} extension is installed and active in your browser`,
+              });
+            }
+          } catch (error) {
+            const errorMessage = await parseError(error);
```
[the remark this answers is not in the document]
elias-garcia: Done!

src/constants.ts
```diff
@@ -35,21 +37,25 @@ export const ETH_TOKEN: Token = {
...
   ethereumRpcUrl: string;
   polygonHermezRpcUrl: string;
+  bridgePolygonHermezNetworkId: number;
 }): [Chain, Chain] => [
   {
     key: "ethereum",
     name: "Ethereum chain",
     Icon: EthChainIcon,
     provider: new JsonRpcProvider(ethereumRpcUrl),
+    bridgeNetworkId: 0,
```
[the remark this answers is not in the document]
elias-garcia: I'll change it based on the decision on the comment above!

src/views/home/components/list/list.view.tsx
elias-garcia: Removed!

src/contexts/providers.context.tsx
elias-garcia: I think that this is related to the latest PR merged. I'll create an issue to take a more detailed look to this in another PR

src/contexts/providers.context.tsx
amonsosanz: Do you mean this "account connection" screen? I'm unable to reproduce the case...

image

amonsosanz: Got it. Here: 
image

src/views/home/components/list/list.view.tsx
```diff
@@ -28,31 +27,9 @@ interface ListProps {
...
-  const [values, setValues] = useState(list);
-
-  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const value = event.target.value.trim();
-    const filtered: List =
-      list.type === "chain" ? filterChainList(list, value) : filterTokenList(list, value);
-
-    setValues(filtered);
-  };
+  const [values] = useState(list);
```
amonsosanz, line 32: If we no longer modify this state we can remove it and use the `list` straight.

elias-garcia: Done!

src/constants.ts
```diff
@@ -35,21 +37,25 @@ export const ETH_TOKEN: Token = {
 export const getChains = ({
   ethereumRpcUrl,
   polygonHermezRpcUrl,
+  bridgePolygonHermezNetworkId,
```
amonsosanz, line 40: We can remove the prefix also here.

src/adapters/env.ts
```diff
@@ -22,11 +23,17 @@ const envToDomain = ({
   REACT_APP_BRIDGE_API_URL,
   REACT_APP_L1_BRIDGE_CONTRACT_ADDRESS,
   REACT_APP_L2_BRIDGE_CONTRACT_ADDRESS,
+  REACT_APP_POLYGON_HERMEZ_NETWORK_ID,
   REACT_APP_FIAT_EXCHANGE_RATES_API_URL,
   REACT_APP_FIAT_EXCHANGE_RATES_API_KEY,
   REACT_APP_USDT_ADDRESS,
   REACT_APP_UNISWAP_QUOTER_CONTRACT_ADDRESS,
 }: Env): domain.Env => {
+  const bridgePolygonHermezNetworkId = z
```
amonsosanz, line 32: We can remove the prefix also here.

src/contexts/bridge.context.tsx
```diff
@@ -130,68 +154,123 @@ const BridgeProvider: FC = (props) => {
...
   );
 
-  const bridge = useCallback(
-    async (
-      token: string,
-      amount: BigNumber,
-      destinationNetwork: number,
-      destinationAddress: string
-    ): Promise<ContractTransaction> => {
+  const estimateGasPrice = useCallback(
```
amonsosanz, line 157: we can remove the `useCallback` right?

src/views/home/components/amount-input/amount-input.view.tsx
```diff
@@ -15,25 +15,26 @@ interface AmountInputProps {
...
+      if (amount && actualFee) {
         const newAmountWithFee = amount.add(actualFee);
         const isNewAmountWithFeeMoreThanFunds = newAmountWithFee.gt(balance);
         const error = isNewAmountWithFeeMoreThanFunds ? "Insufficient balance" : undefined;
-        onChange({ amount, error });
+
+        return onChange({ amount, error });
       } else {
-        onChange({});
+        return onChange({});
```
amonsosanz, line 37: Why add these `return` statements? `onChange` returns nothing...

src/views/home/components/transaction-form/transaction-form.view.tsx
```diff
@@ -14,72 +13,118 @@ import Button from "src/views/shared/button/button.view";
...
+  }, [env, transaction]);
 
   useEffect(() => {
-    if (env && !transaction) {
-      setFormData({
-        from: env.chains[0],
-        to: env.chains[1],
-        token: env.tokens.ETH,
-      });
+    setAmount(undefined);
```
amonsosanz, line 94: The previous effect is in charge of restoring the data from the `TransactionData` present in the context, including the Chains. Whenever the chains are updated, this effect is executed, resetting the amount back to `undefined` and thus breaking the behavior of retrieving the amount entered by the user when the Confirm Transfer screen is closed.

src/contexts/bridge.context.tsx
```diff
@@ -130,68 +154,123 @@ const BridgeProvider: FC = (props) => {
...
         throw new Error("Bridge contract is not available");
       }
 
-      if (token === ethersConstants.AddressZero) {
-        return bridgeContract.bridge(token, amount, destinationNetwork, destinationAddress, {
-          value: amount,
+      return contract.estimateGas
+        .bridge(token.address, amount, to.networkId, destinationAddress, {
+          ...overrides,
+          from: destinationAddress,
```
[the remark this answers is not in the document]
amonsosanz: I now realize that indeed [you pass the `account` as `destinationAddress`](https://github.com/hermeznetwork/bridge-ui/pull/54/files#diff-11deaa099034e63533a095b4f47018bb938bca4fab21c9c2d6f011108f9cfc35R107). This is a bit messy, to be honest.

It feels like as `destinationAddress` we should pass the actual `destinationAddress` of the bridge, so we can simulate the real transaction.

Then we may miss a param `originAddress` to (if required) be passed as the originating address of the transaction via the overrides.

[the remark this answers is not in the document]
amonsosanz: Ok... [we always assume the `destinationAddress` to be the same address as the user's address](https://github.com/hermeznetwork/bridge-ui/pull/54/files#diff-e1a070e92493df886cfab4008e0eaf0149bea35813d1368ac30bd78380d1ddd4R39), aka `account`.

It's the name of the param that is misleading me. If we're using the same address as origin and destination, this param should probably be named just address in the domain of the bridge.

Do we plan to enable the feature of creating a bridge to be claimed by a different address?

src/constants.ts
elias-garcia: Done!

src/adapters/env.ts
elias-garcia: Done!

src/contexts/bridge.context.tsx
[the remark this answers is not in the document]
elias-garcia: Yes, the `destinationAddress` is the one who is going to receive the funds. In the `bridge-ui` we are going to support only bridges to yourself. For me the name makes sense, as it's the destination address which is going to receive the funds. Take into account that when we retrieve the pending `bridges` from the UI, we are retrieving them by `destinationAddress`, so if someone call the smart contract directly with another address as destination, this are going to appear as well in the UI and the user should be able to claim them. So just `address` is not complete.

src/views/home/components/amount-input/amount-input.view.tsx
elias-garcia: I like to be explicit on what the function returns. It's more clear to read and to understand what a function does just by taking a quick look to it.

src/contexts/bridge.context.tsx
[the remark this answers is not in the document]
amonsosanz: Ok, that makes sense, although reading `from: destinationAddress` is a bit weird. One would expect to read `from: originAddress` (or `to: destinationAddress`). So I'd prefer passing both `destinationAddress` and `originAddress` to this function. If it turns out that they are now the same address, nice, no prob with that. This make explicit what you were commenting:

> In the `bridge-ui` we are going to support only bridges to yourself.

By the way, I still would like to know why is this `from` param required in overrides.

src/views/home/components/amount-input/amount-input.view.tsx
amonsosanz: > I like to be explicit on what the function returns.

I like to be explicit too, the point is that this function returns nothing. Without the `return`, this was clear. Now with `return onChange()` it's no longer explicit. I need to go to the signature of `onChange` to see if I'm returning something from `updateAmountInput`.

With this change, you made the explicit implicit.

amonsosanz: We agreed on leaving it since the return makes explicit when the function ends.

amonsosanz (approved):

PR #55, opened 2022-04-26
Refine bridge domain and implement activity list

amonsosanz:
This PR unifies the bridge domain types `Bridge`, `Claim`, `ClaimStatus`, and `MerkleProof` on a single `Transaction` entity and implements a function `getTransactions` in adapters which takes care of doing all the requests and building the `Transaction` list.

It also removes the faked data from the `Activity` view and lists the transactions, enabling the claim against the smart contract when possible (the `ClaimStatus` for the `Bridge` is ready and the `Claim` does not yet exist).


src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -1,37 +1,109 @@
...
+            });
+          });
+        }
+      });
+    }
+  };
+
+  useEffect(() => {
+    if (account.status === "successful") {
+      // ToDo: Get all the data only for the right bridge
```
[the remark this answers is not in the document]
amonsosanz: Currently `getTransactions` is getting all the `bridges/claims` and per each bridge, the `proof` and the `claim-status`, while here, we only need to get the `proof` and the `claim-status` of the affected bridge/claim pair.

src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -42,48 +114,40 @@ const TransactionDetails: FC = () => {
...
+          <Chain
+            chain={destination.networkId === 0 ? "ethereum" : "polygon"}
+            className={classes.alignRow}
+          />
         </div>
-        <div className={classes.row}>
+        {/* <div className={classes.row}>
           <Typography type="body2" className={classes.alignRow}>
             L2 Fee
           </Typography>
```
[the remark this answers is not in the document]
amonsosanz: Done

src/views/activity/components/transaction-card/transaction-card.styles.ts
```diff
@@ -76,7 +76,7 @@ const useTransactionCardStyles = createUseStyles((theme: Theme) => ({
     justifyContent: "space-between",
     alignItems: "center",
   },
-  finaliseButton: {
+  finalizeButton: {
```
elias-garcia, line 79: `finalise` is more correct than `finalize`. This was corrected by Jenna in the wallet.

src/domain/index.ts
```diff
@@ -88,20 +71,62 @@ export type Message =
       parsed: string;
     };
 
+export function getTransactionStatusText(status: Transaction["status"]): string {
```
elias-garcia, line 74: How we display the text in the UI shouldn't be part of the domain. If you need to reuse this, it should be a component: `TransactionStatus`.

src/domain/index.ts
```diff
@@ -88,20 +71,62 @@ export type Message =
...
+    case "initiated":
+      return "Initiated";
+    case "on-hold":
+      return "On Hold";
+    case "completed":
+      return "Completed";
+  }
+}
+
+export function getChainName(chain: Chain): string {
```
elias-garcia, line 85: Same as before, this shouldn't be in the domain. If you need to display it across different parts of the application, it should be a component: `ChainLabel` or `ChainName`.

src/views/home/components/list/list.view.tsx
```diff
@@ -41,27 +41,27 @@ const List: FC<ListProps> = ({ list, onClose }) => {
...
           <div className={classes.list}>
             {list.type === "chain"
-              ? list.items.slice(0, 20).map((element) => {
+              ? list.items.slice(0, 20).map((chain) => {
                   return (
                     <button
                       className={classes.button}
-                      key={element.name}
-                      onClick={() => list.onClick(element)}
+                      key={getChainName(chain)}
```
elias-garcia, line 48: I don't like this, we should use unique identifiers for the `key` and a label is not the best idea. If we use `chain.chainId` we can guarantee that this key is unique.

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -1,45 +1,34 @@
...
-  const env = useEnvContext();
-  const LayerIcon = status !== "completed" && status !== "failed" ? ReloadIcon : layerIcons[target];
-  const actionText = target === "l1" ? "Transfer to L1" : "Transfer to L2";
+  const LayerIcon =
+    status !== "completed"
+      ? ReloadIcon
+      : destinationNetwork.key === "ethereum"
+      ? TransferL1Icon
+      : TransferL2Icon;
+  const actionText = destinationNetwork.key === "ethereum" ? "Transfer to L1" : "Transfer to L2";
```
elias-garcia, line 30: I would render both `LayerIcon` and `actionText` directly in the `JSX` as they are only used once.

src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -54,43 +43,39 @@ const TransactionCard: FC<TransactionCardProps> = ({
...
+          </span>
         </div>
         <div className={classes.tokenColumn}>
           <div className={classes.token}>
-            {env && <Icon url={env.tokens.ETH.logoURI} className={classes.tokenIcon} size={20} />}
-            <Typography type="body1">
-              {amount} {token.toUpperCase()}
-            </Typography>
+            <Icon url={transaction.token.logoURI} className={classes.tokenIcon} size={20} />
+            <Typography type="body1">{formatEther(amount)} ETH</Typography>
```
elias-garcia, line 55: Don't use `formatEther`, be more generic and do: `${formatUnits(amount, token.decimals)} ${token.symbol}`. This will support ERC-20 tokens in the future.

src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -1,37 +1,110 @@
...
+
   return (
     <>
       <Header title="Transaction Details" backTo="activity" />
       <Card className={classes.card}>
         <div className={classes.balance}>
-          {env && <Icon url={env.tokens.ETH.logoURI} className={classes.tokenIcon} size={48} />}
-          <Typography type="h2">{`${amount} ${token.toUpperCase()}`}</Typography>
+          <Icon url={token.logoURI} className={classes.tokenIcon} size={48} />
+          <Typography type="h2">{`${formatEther(amount)} ETH`}</Typography>
```
elias-garcia, line 107: Same as before, `${formatUnits(amount, token.decimals) ${token.symbol}}`

src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -95,9 +138,13 @@ const TransactionDetails: FC = () => {
...
         <div className={classes.finaliseRow}>
-          <button className={classes.finaliseButton} disabled={status === "initiated"}>
+          <button
+            onClick={onClaim}
+            className={classes.finaliseButton}
+            disabled={status === "initiated"}
+          >
             <Typography type="body1" className={classes.finaliseButtonText}>
-              Finalise
+              Finalize
```
elias-garcia, line 147: Same as before, `Finalise` was correct.

src/views/transaction-details/transaction-details.view.tsx
```diff
@@ -1,37 +1,110 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if (account.status === "successful") {
+      // ToDo: Get all the data only for the right bridge
+      void getTransactions({ ethereumAddress: account.data })
+        .then((transactions) => {
+          const foundTransaction = transactions.find((tx) => {
+            const id = `${tx.destinationNetwork.networkId}-${tx.depositCount}`;
```
elias-garcia, line 65: I see that you use a lot here and there:

```
const id = `${tx.destinationNetwork.networkId}-${tx.depositCount}`;
```

Why not adding it to the `Transaction` domain when calling `getTransactions`? At least doing this we'll have the `id` centralized in one point. It's safer If we need to change it in the future as well.

src/views/activity/activity.view.tsx
```diff
@@ -1,22 +1,73 @@
...
+            openSnackbar({
+              type: "error",
+              parsed,
+            });
+          });
+        });
+    }
+  }, [account, getTransactions, openSnackbar]);
+
+  const pendingTransactions = transactionList.filter((data) => data.status !== "completed");
```
elias-garcia, line 68: I would move this `const` to the beginning of the file (to where all the constants are).

src/adapters/bridge-api.ts
```diff
@@ -28,139 +27,205 @@ interface MerkleProof {
...
+        destinationNetwork,
+        networkId,
+        amount,
+        destinationAddress,
+        depositCount,
+      } = bridge;
+      const [claimStatus, merkleProof]: [domain.ClaimStatus, domain.MerkleProof] =
+        await Promise.all([
+          getClaimStatus({ apiUrl, networkId, depositCount }),
+          getMerkleProof({ apiUrl, networkId, depositCount }),
```
elias-garcia, line 174: We shouldn't get the `MerkleProof` until the user does a `claim`.

Doing this, we will request the merkle proof even if the `getClaimStatus` returns `false`. I would request the proof only when the user wants to click on the `Finalise` button to avoid unnecessary calls to the API.

At least, I would request it only when the `getClaimStatus` returns `true`, but this doesn't guarantee that the user is going to use the proof at all (as he only needs it for claiming).

elias-garcia (changes requested):

src/views/activity/components/transaction-card/transaction-card.styles.ts
amonsosanz: Really? I thought it was the other way around.

Reverted!

src/views/home/components/list/list.view.tsx
amonsosanz: Your right, I replaced `element.name` without realizing this should have been `element.key` in the first place.

src/views/activity/activity.view.tsx
amonsosanz: Done

src/views/activity/components/transaction-card/transaction-card.view.tsx
amonsosanz: Done

src/views/transaction-details/transaction-details.view.tsx
amonsosanz: Yeah much better. Done.

src/views/activity/components/transaction-card/transaction-card.view.tsx
amonsosanz: Right. Done!

src/views/transaction-details/transaction-details.view.tsx
amonsosanz: Right. Done!

src/domain/index.ts
amonsosanz: > How we display the text in the UI shouldn't be part of the domain.

Agreed.

> If you need to reuse this, it should be a component: `TransactionStatus`.

I don't agree with this. I think that having a util function as now is more versatile. With a util function, we can get the text and use it to build a msg for the user, show a log, or anything else, while a component limits the usage to JSX.

amonsosanz: Done. Both utils moved to a new module `utils/labels.ts`.

src/domain/index.ts
amonsosanz: Done. Both utils moved to a new module `utils/labels.ts`.

src/adapters/bridge-api.ts
amonsosanz: Yep, this makes a lot of sense. To be honest I focused on making the thing work and didn't even think about optimizing the flow of requests.

I've removed the `MerkleProof` from the already claimed TXs (`status: "completed"`) since once a TX has been claimed we no longer need to get the `MerkleProof`. This alone will prevent obtaining the `MerkleProof` for all the historic TXs.

Also in `getTransactions` now the first thing we do is check whether the TX has been claimed. If a TX has been already claimed we can return right away without even having to call `claimStatus`.

If the TX can't be found in the claim list, then we call `getClaimStatus` and return right away when the claim is not yet ready.

And finally, only when the claim is ready we do call `getMerkleProof` and return a TX with `status: "on-hold"`.

Given that the list of TXs pending to be claimed is expected to be small and that the `MerkleProof` is a lightweight payload, I think this level of optimization is more than enough (i.e. I'd not obtain the `MerkleProof` when the user clicks the Claim button)

adrifdez (approved):

PR #63, opened 2022-05-03
Add multiplatform docker build workflow and tag images for integratio…

xavier-romero:
### What does this PR does?
Introduce GitHub workflows to build multi platform docker images that can be directly run also in ARM MacBook computers. Also set new tags for integration purposes: **1.5-integration**

### How to test?
You'll see that an Action is run for every push to develop branch

## Checklist

review them:

@elias-garcia


.gitignore
```diff
@@ -7,6 +7,7 @@ node_modules/
 # Environment
 .env*
 !.env.example
+!.env.integration
```
elias-garcia: I think that now makes more sense just to ignore the `.env`:

```
# Environment
.env
```

I've just pushed a commit with this change :)

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+        # Redirect all requests to index.html \n\
+        try_files $uri /index.html =404; \n\
+    } \n\
+\n\
+    error_page   500 502 503 504  /50x.html; \n\
+    location = /50x.html { \n\
+        root   /usr/share/nginx/html; \n\
+    } \n\
+} \n\
+' > /etc/nginx/conf.d/default.conf
```
elias-garcia, line 36: Instead of hardcoding this code, I would suggest to create a `deployment` folder in the root of the project and put an `nginx.conf` file inside (`/deployment/nginx.conf`).

Then you can just do:

```
RUN mv deployment/nginx.conf /etc/nginx/conf.d/default.conf
```

I think that it's better, specially if we need tom modify the nginx config, but as you prefer Xavi!

elias-garcia (approved):

PR #66, opened 2022-05-05
env for internal-testnet

xavier-romero:
### What does this PR does?

Add the right configuration to build and push Docker images for internal-testnet 1.5 when pushing changes to develop branch.

### How to test?
Push to develop and after some minutes check on dockerhub for a new image with tag 1.5-itestnet


.env.internal-testnet
```diff
@@ -0,0 +1,18 @@
+# ETHEREUM
```
amonsosanz, line 1: We have added 3 new env vars since this PR was opened.

PR #69, opened 2022-05-06
Initialize each Chains with its Network

amonsosanz:
This PR initializes each Chain with its Network.


src/contexts/bridge.context.tsx
```diff
@@ -139,7 +139,10 @@ const BridgeProvider: FC = (props) => {
         throw new Error("Connected provider is not available");
       }
 
-      const contract = Bridge__factory.connect(from.contractAddress, connectedProvider.getSigner());
+      const contract = Bridge__factory.connect(
+        from.contractAddress,
+        connectedProvider.provider.getSigner()
```
elias-garcia, line 144: Why we access `connectProvider.provider` instead of just `connectedProvider` as we were doing before?

src/views/activity/activity.view.tsx
```diff
@@ -28,13 +28,11 @@ const Activity: FC = () => {
 
   const onClaim = async (tx: Transaction) => {
     if (tx.status === "on-hold") {
-      if (!(await isConnectedProviderChainOk(tx.bridge.destinationNetwork))) {
+      if (isConnectedProviderChainOk && !isConnectedProviderChainOk(tx.bridge.destinationNetwork)) {
```
elias-garcia, line 31: I think that `isConnectedProviderChainOk` can't be `null` or `undefined` right?

src/contexts/providers.context.tsx
```diff
@@ -184,54 +181,47 @@ const ProvidersProvider: FC = (props) => {
...
-            }
-            throw new Error("Unknown error from Metamask");
-          }
-        }
-      } else {
-        return Promise.reject(new Error(providersContextNotReadyErrorMsg));
+            params: [{ chainId: hexValue(chain.chainId) }],
+          })
+            // eslint-disable-next-line @typescript-eslint/no-empty-function
+            .then(() => {})
```
elias-garcia, line 198: Just do: `.then(() => ({}))` so we can delete the `eslint-disable-next-line` above.

src/views/activity/activity.view.tsx
adrifdez: Yes, that one keeps as before with await

src/contexts/providers.context.tsx
adrifdez: If we add the parenthesis we'll return and object instead of a function

src/contexts/providers.context.tsx
```diff
@@ -18,7 +18,7 @@ import routes from "src/routes";
 import { Chain, EthereumEvent, WalletName } from "src/domain";
 
 interface ProvidersContext {
-  connectedProvider?: Web3Provider;
+  connectedProvider?: { provider: Web3Provider; chainId: number };
   account: AsyncTask<string, string>;
   isConnectedProviderChainOk: (chain: Chain) => Promise<boolean>;
```
amonsosanz, line 23: This function should not return a promise anymore.
Indeed, now that the chainId is syncronously available, this function should not exist at all. We can replace the calls to this function with:

```typescript
if (connectedProvider && connectedProvider.chainId === chainId) {...}
```

amonsosanz: This should not be like this. I see 2 problems in the current implementation:

* The signature of `changeNetwork` should be: `(chain: Chain) => Promise<number>` and its implementation should change the network and then return the `chainId` of the newly selected network.
* After the network has been changed, we need to call `setConnectedProvider` and update the value of `chainId` to the new selected network.

Sadly, I see that MetaMask resolves to `null` in both `wallet_switchEthereumChain` and `wallet_addEthereumChain` after changing the `Network`, so our implementation should retrieve the network after the change to be able to return it and store it in the `connectedProvider` state.

I'd also suggest to split the function `changeNetwork` into 3 functions like this to have a clear separation of concerns:

```typescript
const switchNetwork = (chain: Chain, connectedProvider: Web3Provider): Promise<number> => {
  if (connectedProvider.provider.request) {
    return connectedProvider.provider
      .request({
        method: "wallet_switchEthereumChain",
        params: [{ chainId: hexValue(chain.chainId) }],
      })
      .then(async () => {
        const { chainId } = await connectedProvider.getNetwork(); // we retrieve the current network
        setConnectedProvider({ provider: connectedProvider, chainId: chainId }); // update our state
        return Promise.resolve(chainId); // and return the new value
      });
  }
  return Promise.reject(new Error("The provider does not have a request method"));
};

const addAndSwitchNetwork = (chain: Chain, connectedProvider: Web3Provider): Promise<number> => {
  if (connectedProvider.provider.request) {
    return connectedProvider.provider
      .request({
        method: "wallet_addEthereumChain",
        params: [
          {
            chainId: hexValue(chain.chainId),
            chainName: getChainName(chain),
            rpcUrls: [chain.provider.connection.url],
          },
        ],
      })
      .then(async () => {
        const { chainId } = await connectedProvider.getNetwork(); // we retrieve the current network
        setConnectedProvider({ provider: connectedProvider, chainId: chainId }); // update our state
        return Promise.resolve(chainId); // and return the new value
      });
  }
  return Promise.reject(new Error("The provider does not have a request method"));
};

const changeNetwork = useCallback(
  (chain: Chain) => {
    if (connectedProvider && connectedProvider.provider.provider.isMetaMask) {
      return switchNetwork(chain, connectedProvider.provider).catch((error) => {
        if (isMetamaskUnknownChainError(error)) {
          return addAndSwitchNetwork(chain, connectedProvider.provider);
        } else {
          throw error;
        }
      });
    } else {
      return Promise.reject(new Error(providersContextNotReadyErrorMsg));
    }
  },
  [connectedProvider]
);
```

src/views/activity/activity.view.tsx
elias-garcia: But why do you check if `isConnectedProviderChainOk` exists then? And why are not we awaiting for the result as before if the function still returns a `Promise`?

src/contexts/providers.context.tsx
```diff
@@ -165,73 +161,66 @@ const ProvidersProvider: FC = (props) => {
...
+      return connectedProvider.provider
+        .request({
+          method: "wallet_switchEthereumChain",
+          params: [{ chainId: hexValue(chain.chainId) }],
+        })
+        .then(async () => {
+          const { chainId } = await connectedProvider.getNetwork();
+          return chainId === chain.chainId
+            ? chainId
+            : Promise.reject("Could not switch the network");
```
amonsosanz, line 175: We should reject with an Error. Same below.

PR #80, opened 2022-05-13
Rename routes

### LINKED ISSUE #13: Design routing
We currently have a route `/transaction-details/:transactionId`, which feels a bit awkward. Probably `/transactions/:transactionId` would feel better, but it's not the case considering that we also have the route `/activity`, which lists all transactions.

Shouldn't `transaction-details` be a nested route of `/activity`? Shouldn't we consider renaming `/activity` to `/transactions`?

This issue is created to have a space to discuss and design the routes of the application.

elias-garcia:
Closes #13.

### What does this PR does?

This PR renames the routes of the app and the top views used by the routes.

## Checklist

review them:


src/views/activity/components/transaction-card/transaction-card.view.tsx
```diff
@@ -2,8 +2,8 @@ import { FC } from "react";
 import { useNavigate } from "react-router-dom";
 
 import useTransactionCardStyles from "src/views/activity/components/transaction-card/transaction-card.styles";
```
adrifdez, line 4: This component should change to BridgeCard

elias-garcia: Right, done! I've also renamed the `TransactionForm` to `BridgeForm`.

adrifdez (approved):

amonsosanz (approved):

src/contexts/form.context.tsx
```diff
@@ -0,0 +1,32 @@
...
+    console.error("The form context is not yet ready");
+  },
+};
+
+const formContext = createContext<FormContext>(formContextDefaultValue);
+
+const FormProvider: FC = (props) => {
+  const [formData, setFormData] = useState<FormData>();
+
+  const value = useMemo(() => {
```
adrifdez, line 21: We don't need useMemo to check a state value

PR #84, opened 2022-05-18
Remove list abstraction

amonsosanz:
This PR removes the abstraction of the `List` component, creating 2 components: `ChainList` and `TokenList`.


src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import useListStyles from "src/views/home/components/token-list/token-list.styles";
+import Card from "src/views/shared/card/card.view";
+import Typography from "src/views/shared/typography/typography.view";
+import Icon from "src/views/shared/icon/icon.view";
+import Portal from "src/views/shared/portal/portal.view";
+import { Token } from "src/domain";
+
+interface TokenListProps {
+  placeholder: string;
```
adrifdez, line 11: Why do we need a placeholder?

amonsosanz: we don't!

amonsosanz: Well, indeed, I remove it from `ChainList` and somehow leave it here because we will have an input. Do you think it may make sense to leave a hint like "Contract address"?

amonsosanz: Well, in this PR, I'm removing it. We can add it later. Done!

adrifdez (approved):

PR #92, opened 2022-05-24
Only show tokens that belong to the network selected in from

amonsosanz:
This PR changes the behavior of a [previous PR](https://github.com/hermeznetwork/bridge-ui/pull/85) by only listing the tokens that belong to the network selected in the `from` input.


src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -55,19 +56,28 @@ const BridgeForm: FC<BridgeFormProps> = ({ account, formData, resetForm, onSubmi
     status: "pending",
   });
 
+  const getEtherToken = (chain: Chain): Token | undefined => {
+    return getChainTokens(chain).find((token) => token.address === ethers.constants.AddressZero);
+  };
```
amonsosanz, line 61: We could remove this function and use `constants.getEtherToken` directly, but this feels more correct.

elias-garcia (approved):

PR #94, opened 2022-05-25
Redirect to the log in page when a non-supported network is selected in MetaMask

### LINKED ISSUE #90: Grey screen when switching to a wrong network
When you are in the bridge form and you switch your network in MetaMask to a non supported network, the form disappears and you only see a grey screen:

Captura de Pantalla 2022-05-24 a las 14 04 04

We should detect this and redirect the user to the login screen.

amonsosanz:
Closes #90

### What does this PR does?

This PR makes the application navigate to the login page whenever a non-supported network is selected in MetaMask.

### DIFF
--- a/src/contexts/providers.context.tsx
+++ b/src/contexts/providers.context.tsx
@@ -222,6 +222,12 @@ const ProvidersProvider: FC = (props) => {
   );
 
   useEffect(() => {
+    if (account.status === "failed") {
+      navigate(routes.login.path);
+    }
+  }, [account, navigate]);
+
+  useEffect(() => {
     const internalConnectedProvider: Record<string, unknown> | undefined =
       connectedProvider?.provider.provider;
     const onAccountsChanged = (accounts: unknown): void => {
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -12,27 +12,27 @@ const routes = {
     isPrivate: true,
   },
   login: {
-    path: "/login",
+    path: "/login" as const,
     Component: Login,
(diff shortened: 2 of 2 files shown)


src/contexts/providers.context.tsx
```diff
@@ -221,6 +221,12 @@ const ProvidersProvider: FC = (props) => {
     [connectedProvider]
   );
 
+  useEffect(() => {
+    if (account.status === "failed") {
```
elias-garcia, line 225: I think that the account can also has the status `failed` in the login page, wouldn't this trigger an infinite loop or a refresh?

amonsosanz: Not sure I follow... the account's state is only stored in the provider. What we do on the log-in page is reset the state of `selectedWallet` when the account is `undefined`.

With this addition, when `account.status === "failed"` we will do 2 things:
* In `login.view.tsx` we reset the `selectedWallet`, which is ok since the state of the selected wallet is managed there.
* In `providers.context.tsx` we redirect to the path `/login` so if we were not yet there, we go there and do nothing otherwise.

What could cause an infinite loop?

amonsosanz: > What happens if we are in the `login` page and we redirect again to the `login`?

In my initial implementation, I was checking the current location and only redirecting to login when we were not on the login page. Later I confirmed that the router does nothing when we redirect to login from login so I removed this condition. From my tests, it works ok.

elias-garcia (approved):

PR #105, opened 2022-06-08
Fix react setState errors when components unmount

### LINKED ISSUE #101: Fix React error when component unmounts
image-3

amonsosanz:
Closes #101.

### What does this PR does?

This PR prevents unmounted components from making changes to their state after asynchronous processes (promise resolution) have finished.

After this PR is merged these errors should not appear anywhere under any circumstance:

image


src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -54,13 +56,22 @@ const BridgeForm: FC<BridgeFormProps> = ({ account, formData, resetForm, onSubmi
...
   const [inputError, setInputError] = useState<string>();
-  const [chains, setChains] = useState<FormChains>();
+  const [chains, setChains] = useState<Chains>();
   const [token, setToken] = useState<Token>();
   const [amount, setAmount] = useState<BigNumber>();
   const [estimatedFee, setEstimatedFee] = useState<AsyncTask<BigNumber, string>>({
     status: "pending",
   });
 
+  const mountSafe = useCallback(
```
elias-garcia, line 66: I don't like having to copy and paste this here and there... Can't we create a hook that internally uses the `useIsMounted`?

Apart from that, why instead calling the `mountSafe` function as:

```ts
mountSafe(notifyError, error);
```

we don't just send a callback like this?

```ts
mountSafe(() => notifyError(error));
```

It looks more readable and it's a more familiar javascript code.

amonsosanz: > `mountSafe(() => notifyError(error));`

Yep, much better, didn't think about that option. Closures to the rescue.

amonsosanz: > Can't we create a hook that internally uses the `useIsMounted`?

Done

PR #107, opened 2022-06-10
Fix or add the loading spinners as required

### LINKED ISSUE #96: Fix or add the loading spinners as required
The objective of this task is to add all the loaders that are necessary and fix those that currently do not work well or are not properly positioned on the screen.

Here an example of a broken loader:

image

amonsosanz:
Closes #96.

### What does this PR does?

This PR makes sure that all pages that require it show a spinner while the data is being fetched.

### DIFF
--- a/src/views/activity/activity.view.tsx
+++ b/src/views/activity/activity.view.tsx
@@ -85,12 +85,16 @@ const Activity: FC = () => {
   useEffect(() => {
     setWrongNetworkBridges([]);
   }, [connectedProvider?.chainId]);
-
   switch (bridgeList.status) {
     case "pending":
     case "loading":
     case "reloading": {
-      return <PageLoader />;
+      return (
+        <>
+          <Header title="Activity" backTo="home" />
+          <PageLoader />
+        </>
+      );
     }
     case "failed": {
       return (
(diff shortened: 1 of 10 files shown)


src/views/activity/activity.view.tsx
```diff
@@ -82,38 +85,65 @@ const Activity: FC = () => {
...
+          <Header title="Activity" backTo="home" />
+          <PageLoader />
+        </>
+      );
+    }
+    case "failed": {
+      return (
+        <>
+          <Header title="Activity" backTo="home" />
+          <Error error={bridgeList.error} />
```
elias-garcia, line 103: Should we display another error along the snackbar?

amonsosanz: I thought that an error was better than nothing or a redirect. What would you show?

elias-garcia: Maybe I would just keep display the spinner and the snackbar. Displaying the same error twice in the screen and in the snackbar looks quite redundant.

amonsosanz: mmm... a spinner indicates that a process is still in progress, while here, we have an error-ended process. I prefer showing something finale. I agree though that the double error feels redundant (although one could close the snackbar and still have the persistent error):

image

What about a no-data message in the line of "No bridges available"? This would also be shown when the list is empty and no error occurred (initial state or empty due to the filter).

PR #108, opened 2022-06-16
Add pagination to the bridge list in Activity

### LINKED ISSUE #104: Add pagination to the Bridge list
Currently the API has a default pagination of 25 items. We need to add support for pagination to be able to display more than 25 items. We can reuse the `InfiniteScroll` component from the wallet.

amonsosanz:
Closes #104

### What does this PR does?

This PR implements the new pagination API of the bridge's service.

Initially, only the first 25 elements are loaded and from there, more batches/pages are loaded as required when the user scrolls to the bottom of the page or uses the "Load more" button.

The polling has been refactored to exactly reload the number of bridges visible, if required with several requests, in batches of 100 items.

### DIFF
--- a/src/adapters/bridge-api.ts
+++ b/src/adapters/bridge-api.ts
@@ -3,6 +3,7 @@ import { z } from "zod";
 
 import { StrictSchema } from "src/utils/type-safety";
 import * as domain from "src/domain";
+import { PAGE_SIZE } from "src/constants";
 
 interface DepositInput {
   token_addr: string;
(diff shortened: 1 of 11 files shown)


src/contexts/bridge.context.tsx
```diff
@@ -373,6 +402,59 @@ const BridgeProvider: FC = (props) => {
...
+                cancelToken: refreshCancelTokenSource.current.token,
+              });
+            })
+        )
+      ).reduce((acc, curr) => [...acc, ...curr], []);
+    },
+    [getBridges]
+  );
+
+  const fetchBridges = useCallback(
```
elias-garcia, line 436: What is the point of this `fetchBridges` function instead of calling `getBridges` and `refreshBridges` from the outside?

src/adapters/bridge-api.ts
```diff
@@ -56,6 +47,12 @@ const depositParser = StrictSchema<DepositInput, DepositOutput>()(
     dest_addr: z.string(),
     deposit_cnt: z.string().transform((v) => z.number().nonnegative().parse(Number(v))),
     tx_hash: z.string(),
+    claim_tx_hash: z
+      .string()
+      .transform((v) => (v.length === 0 ? null : v))
```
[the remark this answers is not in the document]
amonsosanz: [Nope](https://app.slack.com/client/T8VPLGV4Z/C02FMUW1AEM/thread/C02FMUW1AEM-1655120686.922779).

src/contexts/bridge.context.tsx
amonsosanz: Taking care of cancelling the previous requests:

`refreshCancelTokenSource.current.cancel()`

elias-garcia: But we can do the same inside the `refreshBridges` function right?

amonsosanz: mmm... no, we need to do it before calling `getBridges` when `params.type === "load"`.

elias-garcia (approved):

PR #110, opened 2022-06-20
Display tx submitted info message

### LINKED ISSUE #100: Display Tx submitted info message
image-2

amonsosanz:
Closes #100

### What does this PR does?

This PR shows a success message on a snackbar when a bridge or claim transaction is successfully sent.


src/constants.ts
```diff
@@ -22,7 +22,7 @@ export const FIAT_DISPLAY_PRECISION = 2;
 
 export const TOKEN_DISPLAY_PRECISION = 6;
 
-export const SNACKBAR_AUTO_HIDE_DURATION = 5000;
+export const SNACKBAR_AUTO_HIDE_DURATION = 8000;
```
elias-garcia, line 25: I wouldn't change this value, 8s is too much for a snackbar

elias-garcia:
Why are we replacing the error icon? In the Figma I see the older one.

Captura de Pantalla 2022-06-27 a las 12 27 29
a

src/constants.ts
amonsosanz: Even with 8 seconds, [we're violating the WCAG requirement](https://www.w3.org/TR/WCAG21/#enough-time) related to time-based operation from an accessibility perspective.

IMO with 5s, it's difficult to read, considering I don't have any visual or (diagnosed) cognitive issues, and I know when and where the snackbar will appear. I feel a user needs 2s just to cognitively context-switch and start reading the text, which leaves 3 seconds for the content.

Anyway, this is not an exact science. I will revert it but wanted to share my reasoning to increase it in the first place.

amonsosanz: Reverted.

amonsosanz:
> Why are we replacing the error icon? In the Figma I see the older one.
> 
> Captura de Pantalla 2022-06-27 a las 12 27 29

I changed the (sharped) error icon to the new (rounded) icon used in the wallet because I added the _success_ and _info_ icons also from the wallet and felt that they looked more consistent (rounded):

image

Possibly I should have asked the design team for the _info_ and _success_ icons but it seemed like a minor change.

elias-garcia (approved):

PR #116, opened 2022-06-28
Support the new API and improve the Bridge Details request

### LINKED ISSUE #112: Support the new API and improve Bridge Details request
The API has been updated to enable a new endpoint that retrieves the complete details of a deposit. The new enpoint replaces the previous endpoint `/claim-status?net_id=<number>&deposit_cnt=<number>`

Taking advantage of this new endpoint we can now stop requesting the first 100 bridges and filtering by the requested one in Brodge Details. With the new changes in the API we can optimize this to request just the desired one.

amonsosanz:
Closes #112 

### What does this PR does?

This PR fixes the application by adding support to the last changes present in the Bridge API and makes the Bridge Details screen request only the required Bridge.


src/contexts/bridge.context.tsx
```diff
@@ -250,6 +260,137 @@ const BridgeProvider: FC = (props) => {
...
+            tokenOriginNetwork: orig_net,
+            claimTxHash: claim.txHash,
+          };
+        }
+      }
+    },
+    [getTokenPrice, getToken]
+  );
+
   const getBridges = useCallback(
```
elias-garcia, line 394: I think that the conversion from `apiDeposit` to `Deposit` is the same as in the `getBridge` function and we can abstract this a bit more later right?

elias-garcia (approved):

PR #124, opened 2022-07-04
Improve list components

### LINKED ISSUE #114: Improve the tokens search bar
Currently, the token list search bar is not aligned with the designs. For example, it's missing the search icon and the clear button is not very aligned with the other icons in the UI (it's also missing padding).

Also, if you try to search for a token address, and then you try to select with the mouse this address but you end up with the mouse outside of the popup it gets closed. The popup should be closing only when you click outside of the popup.

elias-garcia:
Closes #114.

### What does this PR does?

This PR improves the styles and the behavior of the TokenList and ChainList components.

## Checklist

review them:


src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -42,20 +46,43 @@ const TokenList: FC<TokenListProps> = ({
...
+            <Typography type="h2">Select token</Typography>
+            <button className={classes.closeButton} onClick={onClose}>
+              <XMarkIcon className={classes.closeButtonIcon} />
+            </button>
+          </div>
+          <div className={classes.searchInputContainer}>
+            <MagnifyingGlassIcon className={classes.searchIcon} />
+            <input
+              ref={inputRef}
+              placeholder="Enter token name or address"
```
amonsosanz, line 69: Probably not very relevant but we also filter by the symbol of the token.

amonsosanz (approved):

PR #133, opened 2022-08-01
Vite

elias-garcia:
### What does this PR does?

This PR replaces CRA with Vite to improve the developer experience.

## Checklist

review them:


amonsosanz:
why add `public/vite.svg`?

src/main.tsx
```diff
@@ -1,19 +1,18 @@
...
+import ReactDOM from "react-dom/client";
 import { ThemeProvider } from "react-jss";
 import { BrowserRouter } from "react-router-dom";
 import "normalize.css/normalize.css";
 
 import theme from "src/styles/theme";
 import App from "src/views/app.view";
 
-ReactDOM.render(
+ReactDOM.createRoot(document.getElementById("root") as HTMLElement).render(
```
amonsosanz, line 10: I'd prefer null checking over the assertion.

tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
+    "useDefineForClassFields": true,
     "lib": ["dom", "dom.iterable", "esnext"],
-    "allowJs": true,
-    "skipLibCheck": false,
+    "allowJs": false,
+    "skipLibCheck": true,
```
amonsosanz, line 7: Why change this? We should not skip the lib check.

elias-garcia: In fact, we should. There are a lot of type errors in underlying libraries that we cannot solve.

.gitignore
```diff
@@ -1,17 +1,26 @@
...
 # Environment
 .env
 
 # Artifacts
-build/
+dist
 
 # System files
-**/.DS_Store
+.DS_Store
```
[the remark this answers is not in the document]
elias-garcia: Right, fixed!

src/main.tsx
elias-garcia: Done!

tsconfig.json
amonsosanz: The problem with enabling `skipLibCheck` is that we instruct the TypeScript compiler to skip type checking for *all* declaration (.d.ts) files, including our own `vite-env.d.ts`, which can lead to runtime errors.

> There are a lot of type errors in underlying libraries that we cannot solve.

We have always had this enabled in the wallet and in the bridge and never faced the issue you mention with a third-party lib.

index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="description" content="Simple user interface to bridge ETH and your favorite ERC-20 tokens from Ethereum to the Polygon zkEVM and back" />
```
amonsosanz, line 6: This line is not properly formatted.

tsconfig.json
amonsosanz: The problem with enabling `skipLibCheck` is that we instruct the TypeScript compiler to skip type checking for *all* declaration (.d.ts) files, including our own `vite-env.d.ts`, which can lead to runtime errors.

> There are a lot of type errors in underlying libraries that we cannot solve.

We have always had this enabled in the wallet and in the bridge and never faced the issue you mention with a third-party lib.

amonsosanz: No, [we can't](https://www.typescriptlang.org/tsconfig#exclude).

I see that this is indeed an issue with vite. Reverting `skipLibCheck` to `false` a lot of errors appear.

Also, including `src/vite-env.d.ts` does not seem to work 🤔

amonsosanz: pfff... https://github.com/microsoft/TypeScript/issues/30511

src/routes.ts
```diff
@@ -12,27 +12,27 @@ const routes = {
     isPrivate: true,
   },
   login: {
-    path: "/login" as const,
```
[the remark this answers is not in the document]
elias-garcia: Because we are not using `readonly` properties anywhere in the app (interfaces etc), or are we using it here for another reason?

index.html
elias-garcia: Right, it was excluded from the prettier config. Now it's included and formatted.

.prettierignore
```diff
@@ -1,9 +1,10 @@
...
+scripts
 
-# Except these files
-!src
-!*.json
-
-# And then ignore
-src/types
+# Ignore files
+package-lock.json
```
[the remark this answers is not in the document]
elias-garcia: Right! Fixed

src/routes.ts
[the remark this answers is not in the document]
amonsosanz: The reasoning is not related to readonly but to type inference. The `const assertion` makes the compiler infer the most specific possible type, so in this case, `login.path` would be inferred as `"/login"` instead of `string`. This is nice when you use it later:

amonsosanz (approved):

PR #141, opened 2022-08-08
Show the network required by the environment on the login screen

### LINKED ISSUE #132: Show what network should be used to login
Currently, when you are on the bridge login page you don't know which network you should select to be able to access the application.
This information should be on the screen so that the user knows it before clicking the metamask button.

amonsosanz:
Closes #132

### What does this PR does?

This PR will show the network required by the environment on the login screen.


src/views/login/login.view.tsx
```diff
@@ -40,11 +44,23 @@ const Login: FC = () => {
...
+  if (!env) {
+    return null;
+  }
+
+  const deploymentName = getDeploymentName(env.chains[0]);
+  const appName = deploymentName ? `Bridge ${deploymentName}` : "Bridge";
+  const networkName = getNetworkName(env.chains[0]);
+  const networkInfo =
+    networkName !== undefined &&
+    networkName !== "Mainnet" &&
```
elias-garcia, line 56: Why not checking again the `chainId` instead of checking again a label? Then we can remove the `as const` from the `getNetworkName` function and just return a string (which makes more sense for a label as well).

.gitignore
```diff
@@ -2,6 +2,7 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+*.mjs
```
elias-garcia, line 5: I wouldn't ignore all `.mjs` files because at the end it's the format of an ES6 module. I think that something like `vite.config.ts.timestamp*` would be better

src/views/login/login.view.tsx
amonsosanz: Well, the function `getNetworkName` exists to abstract its user from knowing about chainId's. If we start comparing chainId's we can remove the function. I think computing the network's name and then deciding which names I want to react to is a better approach. Why don't you like using `as const`? I think it's perfectly ok that a label is inferred as a subset of string literals.

.gitignore
amonsosanz: Done.

src/views/login/login.view.tsx
elias-garcia: But in this case the user won't read the `getNetworkName` function... And for us is better and easier to compare chain ids than calculating a label of the UI (which is calculated based on the chain id) and then comparing against this label...

amonsosanz: As agreed, I created the `EthereumChainId` enum and used chainId to check for the network.

elias-garcia (approved):

PR #149, opened 2022-08-11
Fix loading buttons

### LINKED ISSUE #146: Disable the Bridge button when it is clicked
We need to disable the "Bridge" button or display a spinner instead in the "BridgeConfirmation" screen when the user clicks it and the MetaMask pop up appears. This will prevent the user to create more than one tx request.

This includes the same button in the `BridgeDetails` component and in the `BridgeCard`.

elias-garcia:
Closes #146.

### What does this PR does?

This PR disables the "Bridge" and "Claim" buttons" where needed. It also fixes some design issues.

## Checklist

review them:


amonsosanz:
This PR is replaced by #155.

PR #150, opened 2022-08-18
Remove the form data context

### LINKED ISSUE #142: Remove FormData Context
Currently, we are using the `FormData` context just to store the data shared between the form and the confirmation screen. the idea would be to use query params instead to avoid this extra context.

amonsosanz:
Closes #142 

### What does this PR does?

This PR replaces the form context with the router state to preserve the data of the form.

The following small fixes are also included:
* Rename polygon hermez env vars to polygon zkEVM
* Fix the polling of the activity view when there are no bridges
* Prevent the To chain from appearing clickable


src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -284,29 +297,30 @@ const BridgeForm: FC<BridgeFormProps> = ({ account, formData, resetForm, onSubmi
...
-      setSelectedChains({ from: formData.from, to: formData.to });
-      setToken(formData.token);
-      setAmount(formData.amount);
-      resetForm();
+      const { from, to, token, amount } = formData;
+      setSelectedChains({ from, to });
+      setToken(token);
+      setAmount(amount);
+      // This is an ugly hack but I don't know of any other way to reset the router state ¯\_(ツ)_/¯
+      navigate(routes.home.path, { replace: true });
```
amonsosanz, line 321: I'm open to better ideas!

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -42,7 +43,9 @@ interface SelectedChains {
   to: Chain;
 }
 
-const BridgeForm: FC<BridgeFormProps> = ({ account, formData, resetForm, onSubmit }) => {
+const BridgeForm: FC<BridgeFormProps> = ({ account, onSubmit }) => {
+  const { state: routerState } = useLocation();
```
elias-garcia, line 47: I don't like to use the `useLocation` hook inside a component that it's not a route. This is not the case by the moment, but imagine that you use this component under another route in the app. They parser could fail and an error would appear.

I'll move this to the `Home` component.

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -284,29 +297,30 @@ const BridgeForm: FC<BridgeFormProps> = ({ account, formData, resetForm, onSubmi
     /*
      * Load the default values after the network is changed
      */
-    if (env && connectedProvider && formData === undefined) {
+    if (env && connectedProvider) {
```
elias-garcia, line 300: We have a big race condition between this `useEffect` hook and the one below. In fact, this is working because how hooks are placed. If you try to move this hook below the one below, you'll see that it will not work as it will set the amount to `undefined` just after correctly set the `amount` to the desired value from the `formData`.

We should investigate how can we improve this.

elias-garcia:
I've just figured out that with the current implementation (on which we both agreed), we can keep only the `FormData` between the `Home` and the `BridgeConfirmation` screen. But it seems weird going to the `Settings` for example and seeing the form being reset... I think that we should be consistent.

I'm wondering if it would be better to just keep using the context as we were doing before...

src/views/home/components/bridge-form/bridge-form.view.tsx
amonsosanz: Yep. We already discussed this in the past and found no simple solution. Do you have new ideas?

amonsosanz:
> I've just figured out that with the current implementation (on which we both agreed), we can keep only the `FormData` between the `Home` and the `BridgeConfirmation` screen. But it seems weird going to the `Settings` for example and seeing the form being reset... I think that we should be consistent.
> 
> I'm wondering if it would be better to just keep using the context as we were doing before...

I actually like this behavior. I think we should only keep the state between the form and the confirmation views (both ways), but when the user goes to any other route, the state should be discarded.

amonsosanz:
We have deprecated this change in order to maintain the state of the form between all the views.

PR #155, opened 2022-09-13
Disable the bridge button when it is clicked

### LINKED ISSUE #146: Disable the Bridge button when it is clicked

amonsosanz:
Closes #146.

### What does this PR does?

This PR disables the "Bridge" and "Claim" buttons" where needed. It also fixes some design issues.

### DIFF
--- a/src/views/activity/activity.view.tsx
+++ b/src/views/activity/activity.view.tsx
@@ -33,7 +33,7 @@ const Activity: FC = () => {
   const [lastLoadedItem, setLastLoadedItem] = useState(0);
   const [total, setTotal] = useState(0);
   const [wrongNetworkBridges, setWrongNetworkBridges] = useState<Bridge["id"][]>([]);
-  const [finaliseDisabledBridges, setFinaliseDisabledBridges] = useState<Bridge["id"][]>([]);
+  const [disabledBridges, setDisabledBridges] = useState<Bridge["id"][]>([]);
   const classes = useActivityStyles({ displayAll });
 
   const headerBorderObserved = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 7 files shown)


src/views/activity/activity.view.tsx
```diff
@@ -33,6 +33,7 @@ const Activity: FC = () => {
   const [lastLoadedItem, setLastLoadedItem] = useState(0);
   const [total, setTotal] = useState(0);
   const [wrongNetworkBridges, setWrongNetworkBridges] = useState<Bridge["id"][]>([]);
+  const [finaliseDisabledBridges, setFinaliseDisabledBridges] = useState<Bridge["id"][]>([]);
```
elias-garcia, line 36: Have you tried to keep the state locally on each `BridgeCard` component instead of having a map of the disabled ones in the parent component?

amonsosanz: mmm... the state would be kept since the component would not be re-rendered because the `key={bridge.id}` would not change. Let me try!

amonsosanz: mmm... wait no, we can't do that. The actual problem is that the user may cancel the TX from Metamask and in such a case, we have to reset this state and we don't have access to the `catch` from the `BridgeCard`. Does this make sense to you?

(Sorry for the spam I should probably think a bit before I start answering comments...)

elias-garcia: yep, makes sense! didn't see that part of the code

elias-garcia, line 36: "Finalise disabled bridges" as a name of a variable sounds quite weird (in fact it sounds like an action). I propose to rename it to just `disabledBridges`, simpler and easy to read. At the end it's just a list of ids from bridges. Another alternative could be `disabledBridgeActions`, but I prefer the first one.

elias-garcia (approved):

src/views/activity/activity.view.tsx
amonsosanz: You're right. I read it as "finalise-disabled bridges" but in camel case, it looks like an action. Done.

elias-garcia (approved):

PR #174, opened 2022-10-03
Improve L1 gas estimation

elias-garcia:
### What does this PR does?

This PR improves the L1 fee estimation. We need to increase it the estimated gas price in a 20% to be sure that successive bridge txs are successful. This will be added in the future for L2 as well (once the gas estimation works fast).

## Checklist

review them:

### DIFF
--- a/src/contexts/bridge.context.tsx
+++ b/src/contexts/bridge.context.tsx
@@ -1,7 +1,7 @@
 import {
   BigNumber,
-  ContractTransaction,
   constants as ethersConstants,
+  ContractTransaction,
   PayableOverrides,
 } from "ethers";
 import { parseUnits } from "ethers/lib/utils";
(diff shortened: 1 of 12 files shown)


amonsosanz (approved):

src/contexts/bridge.context.tsx
```diff
@@ -136,8 +137,8 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
     []
   );
 
-  const estimateBridgeGasPrice = useCallback(
-    ({ from, to, token, destinationAddress }: EstimateBridgeGasPriceParams) => {
+  const estimateBridgeGasLimit = useCallback(
+    ({ from, to, token, destinationAddress }: EstimateBridgeFee) => {
```
amonsosanz, line 141: Looks like this function `estimateBridgeGasLimit` is the one that should be named `estimateBridgeFee` since it's the one returning the interface `EstimateBridgeFee`.

Can we rename the current function `estimateBridgeFee` returning a BigNumber to something else?

amonsosanz (approved):

PR #175, opened 2022-10-03
Add pending bridge transitions

### LINKED ISSUE #167: Improve transitions between the different Bridge statuses
Currently we are not handling the gap between when you send a tx to the blockchain and when it's being returned by the API. We need to handle both `Deposit` and `Claim` transitions in the UI displaying some skeletons (pending cards).

elias-garcia:
Closes #167.

### What does this PR does?

This PR adds pending transitions between the different states of the bridges.

## Checklist

review them:


src/utils/serializers.ts
```diff
@@ -1,6 +1,9 @@
 import { z } from "zod";
```
amonsosanz, line 1: Why do we have so many parsers in a module named serializers?

I'm not sure we need this generic serializers module tbh... I'd rather move these serializers next to their respective parsers in the corresponding adapter. Adapter modules are properly interface-segmented and can hold both the parsers and serializers for each interface (bridge-api, browser, local storage, etc).

We can group together each parser/serializer pair or separate them into blocks using comments:

```
// parsers

...

// serializers

...

```

src/contexts/bridge.context.tsx
```diff
@@ -568,14 +588,110 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
     [getBridges, refreshBridges]
   );
 
+  const cleanPendingTxs = useCallback(
+    (bridges: Bridge[]): Promise<void> => {
+      if (!env) {
+        return Promise.resolve();
```
amonsosanz, line 594: Shouldn't we reject with an error here?

src/contexts/bridge.context.tsx
```diff
@@ -568,14 +588,110 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
...
+            if (hasTxBeenReverted(txReceipt)) {
+              return storage.removePendingTx(pendingTx.depositTxHash);
+            }
+          }
+
+          if (Date.now() > pendingTx.timestamp + PENDING_TX_TIMEOUT) {
+            return storage.removePendingTx(pendingTx.depositTxHash);
+          }
+        })
+      ).then();
```
amonsosanz, line 653: :neutral_face:

src/contexts/bridge.context.tsx
```diff
@@ -568,14 +588,110 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
...
+        await cleanPendingTxs(bridges);
+      }
+
+      if (!env) {
+        throw new Error("Env is not available");
+      }
+
+      return Promise.all(
+        storage.getPendingTxs(env).map(async (tx) => {
+          const chain = env?.chains.find((chain) => chain.networkId === tx.networkId);
```
amonsosanz, line 670: `env` is a param that can't be falsy.

src/contexts/bridge.context.tsx
```diff
@@ -568,14 +588,110 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
...
+          const fiatAmount =
+            chain && env.fiatExchangeRates.areEnabled
+              ? await getTokenPrice({ token: tx.token, chain })
```
amonsosanz, line 673: I might be missing something here but I think the `fiatAmount` will only match the token price when the token is a stablecoin that matches the selected currency.

In other words, to get the `fiatAmount` of a token (let's say UNI), shouldn't we use `multiplyAmounts`?

src/contexts/bridge.context.tsx
```diff
@@ -568,14 +588,110 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
...
+              : undefined;
+
+          return {
+            status: "pending",
+            networkId: tx.networkId,
+            from: tx.from,
+            to: tx.to,
+            depositTxHash: tx.depositTxHash,
+            claimTxHash: tx.status === "pending-claim" ? tx.claimTxHash : undefined,
+            token: tx.token,
```
amonsosanz, line 683: We should probably make use of `addWrappedToken` here to make sure it's complete.

src/contexts/bridge.context.tsx
```diff
@@ -92,8 +108,9 @@ interface BridgeContext {
     bridges: Bridge[];
     total: number;
   }>;
-  bridge: (params: BridgeParams) => Promise<ContractTransaction>;
-  claim: (params: ClaimParams) => Promise<ContractTransaction>;
+  getPendingBridges: (bridges?: Bridge[]) => Promise<PendingBridge[]>;
+  bridge: (params: BridgeParams) => Promise<void>;
+  claim: (params: ClaimParams) => Promise<void>;
```
amonsosanz, line 113: Why convert the return type `ContractTransaction` to `void`?

src/views/activity/activity.view.tsx
```diff
@@ -48,19 +49,26 @@ const Activity: FC = () => {
   const onDisplayPending = () => setDisplayAll(false);
 
   const onClaim = (bridge: Bridge) => {
-    setDisabledBridges((current) => [...current, bridge.id]);
     if (bridge.status === "on-hold") {
+      setDisabledBridges([...disabledBridges, bridge.id]);
```
amonsosanz, line 53: Why this change? Isn't this introducing a bug when we have already some bridges disabled?

src/views/activity/activity.view.tsx
```diff
@@ -48,19 +49,26 @@ const Activity: FC = () => {
...
+          }
           openSnackbar({
             type: "success-msg",
-            text: "Transaction successfully submitted.\nThe list will be updated once it is processed.",
+            text: "Transaction successfully submitted.",
           });
         })
         .catch((error) => {
-          setDisabledBridges((current) => current.filter((id) => id !== bridge.id));
+          setDisabledBridges(disabledBridges.filter((id) => id !== bridge.id));
```
amonsosanz, line 71: Why this change? Isn't this introducing a bug when we have already some bridges disabled?

src/views/bridge-details/bridge-details.view.tsx
```diff
@@ -267,9 +271,11 @@ const BridgeDetails: FC = () => {
       return <Navigate to={routes.activity.path} replace />;
     }
     case "successful": {
-      const { status, amount, from, to, token, depositTxHash } = bridge.data;
+      const { status, amount, from, to, token } = bridge.data;
 
-      const bridgeTxUrl = `${from.explorerUrl}/tx/${depositTxHash}`;
+      const bridgeTxUrl = `${from.explorerUrl}/tx/${
+        bridge.data.status === "pending" ? bridge.data.depositTxHash : bridge.data.depositTxHash
```
amonsosanz, line 277: :neutral_face:

src/adapters/ethereum.ts
```diff
@@ -202,6 +203,18 @@ const getErc20TokenEncodedMetadata = async ({
   return defaultAbiCoder.encode(["string", "string", "uint8"], [name, symbol, decimals]);
 };
 
+function isTxMined(tx: TransactionResponse | null): boolean {
+  return tx !== null && tx.blockNumber !== null;
```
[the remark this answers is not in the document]
elias-garcia: I think that we already had this discussion in the wallet-ui, I've just copied the code from there. It's not possible to override it in `.d.ts` file (don't know why), and if you see the docs:

Captura de Pantalla 2022-10-04 a las 17 55 43

Don't think that we need a parser just for this.

src/utils/serializers.ts
```diff
@@ -9,18 +12,91 @@ interface BridgeId {
...
+  networkId: number;
+  from: Chain;
+  to: Chain;
+  timestamp: number;
+  token: Token;
+  amount: BigNumber;
+}
+
+type PendingDepositTxData = {
+  status: "pending-deposit";
```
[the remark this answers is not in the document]
elias-garcia: Agree, done!

src/utils/serializers.ts
```diff
@@ -9,18 +12,91 @@ interface BridgeId {
...
+    token: tokenParser,
+    amount: z.string(),
+  })
+);
+
+const serializedPendingTxParser = StrictSchema<SerializedPendingTx>()(
+  z.union([serializedPendingTxDepositParser, serializedPendingTxClaimParser])
+);
+
 const deserializeBridgeId = (
```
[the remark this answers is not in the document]
elias-garcia: Which util?

src/adapters/storage.ts
```diff
@@ -18,9 +25,6 @@ export function setCurrency(currency: Currency): Currency {
 }
 
 // Custom Tokens
-
-const CUSTOM_TOKENS_KEY = "customTokens";
```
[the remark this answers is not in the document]
elias-garcia: Yep I know, but I prefer to have all the constants in a single file so you easily know which constants you have without the need to investigate further in another modules of the app

src/utils/serializers.ts
```diff
@@ -9,18 +12,91 @@ interface BridgeId {
...
+type PendingDepositTx = CommonPendingTx & PendingDepositTxData;
+
+type PendingClaimTx = CommonPendingTx & PendingClaimTxData;
+
+export type PendingTx = PendingDepositTx | PendingClaimTx;
+
+interface CommonSerializedPendingTx {
+  networkId: number;
+  from: number;
+  to: number;
```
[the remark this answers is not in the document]
elias-garcia: Right, done!

src/contexts/bridge.context.tsx
elias-garcia: Fixed

src/contexts/bridge.context.tsx
elias-garcia: Right, fixed

src/views/activity/activity.view.tsx
```diff
@@ -80,12 +88,26 @@ const Activity: FC = () => {
 
   const processFetchBridgesSuccess = useCallback(
     (bridges: Bridge[]) => {
-      callIfMounted(() => {
-        setLastLoadedItem(bridges.length);
-        setBridgeList({ status: "successful", data: bridges });
+      void getPendingBridges(bridges).then((pendingBridges) => {
+        callIfMounted(() => {
+          const filteredApiBridges = bridges.filter((apiBridge) => {
```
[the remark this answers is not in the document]
elias-garcia: You need to filter the bridges that are still `on-hold` until they are complete, as now a `PendingBridge` exists and should be displayed instead of the bridge from the api.

src/views/activity/activity.view.tsx
elias-garcia: Nope, why would the number of bridges disabled affect the behavior of this `setDisabledBridges`?

src/views/bridge-details/bridge-details.view.tsx
elias-garcia: This view will be improved in next iterations. It has not been taken into account by the moment

src/utils/serializers.ts
```diff
@@ -9,18 +12,91 @@ interface BridgeId {
...
+  token: Token;
+  amount: string;
+}
+
+type SerializedPendingDepositTx = CommonSerializedPendingTx & PendingDepositTxData;
+
+type SerializedPendingClaimTx = CommonSerializedPendingTx & PendingClaimTxData;
+
+export type SerializedPendingTx = SerializedPendingDepositTx | SerializedPendingClaimTx;
+
```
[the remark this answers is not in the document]
elias-garcia: You like to write a lot! While it's easier to read it's easier to make a mistake in one of the interfaces as well. When you do a change in the `Common` interface it'll propagate to all of the ones that are using it.

src/adapters/storage.ts
```diff
@@ -74,6 +79,59 @@ export function removeCustomToken(token: Token): Token[] {
...
+}
+
+export function setPendingTxs(pendingTxs: PendingTx[]): PendingTx[] {
+  setStorageByKey({
+    key: constants.PENDING_TXS_KEY,
+    value: pendingTxs.map(serializePendingTx),
+  });
+
+  return pendingTxs;
+}
```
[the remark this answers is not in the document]
elias-garcia: Mmmm, I don't think so. For me it doesn't make sense to return the `SerializedPendingTx` as this serialization should happen only on this module. Us (the domain) should work always with `PendingTx`, but we shouldn't know about the serialization/deserialization happening in this module.

src/adapters/storage.ts
```diff
@@ -74,6 +79,59 @@ export function removeCustomToken(token: Token): Token[] {
...
+    value: serializedPendingTxs,
+  });
+}
+
+export function getPendingTxs(env: Env): PendingTx[] {
+  const serializedPendingTxs = getStorageByKey({
+    key: constants.PENDING_TXS_KEY,
+    defaultValue: [],
+    parser: z.array(serializedPendingTxParser),
+  });
```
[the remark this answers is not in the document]
elias-garcia: Good catch, fixed!

src/adapters/storage.ts
```diff
@@ -74,6 +79,59 @@ export function removeCustomToken(token: Token): Token[] {
...
+
+  return pendingTxs;
+}
+
+export function addPendingTx(pendingTx: PendingTx): void {
+  const serializedPendingTxs = getSerializedPendingTxs();
+  const serializedPendingTx = serializePendingTx(pendingTx);
+
+  setSerializedPendingTxs([...serializedPendingTxs, serializedPendingTx]);
+}
```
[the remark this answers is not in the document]
elias-garcia: Same as before, I think that the domain should always work with `PendingTx` but not serialized versions specific only to this module, that's why I'm returning a `void` instead of returning a `PendingTx[]` (because we are not using them).

src/adapters/storage.ts
```diff
@@ -74,6 +79,59 @@ export function removeCustomToken(token: Token): Token[] {
...
+
+export function removePendingTx(depositTxHash: string): void {
+  const serializedPendingTxs = getSerializedPendingTxs();
+  const updatedSerializedPendingTxs = serializedPendingTxs.filter(
+    (tx) => tx.depositTxHash !== depositTxHash
+  );
+
+  setSerializedPendingTxs(updatedSerializedPendingTxs);
+}
```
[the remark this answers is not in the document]
elias-garcia: Same as before regarding returning `SerializedPendingTx`.

src/views/activity/activity.view.tsx
elias-garcia: Nope, why would the number of bridges disabled affect the behavior of this setDisabledBridges?

src/contexts/bridge.context.tsx
elias-garcia: Good catch here, probably a leftover from the first attempts... Fixed!

src/contexts/bridge.context.tsx
elias-garcia: Buf, good one! I totally missed this, fixed!

src/contexts/bridge.context.tsx
elias-garcia: Right, done!

src/adapters/ethereum.ts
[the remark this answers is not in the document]
elias-garcia: BTW this have been fixed in ethers v6 (they don't want to update this now because it would break a lot of JS builds :)).

src/adapters/storage.ts
```diff
@@ -74,6 +79,59 @@ export function removeCustomToken(token: Token): Token[] {
...
+
+export function getPendingTxs(env: Env): PendingTx[] {
+  const serializedPendingTxs = getStorageByKey({
+    key: constants.PENDING_TXS_KEY,
+    defaultValue: [],
+    parser: z.array(serializedPendingTxParser),
+  });
+
+  return serializedPendingTxs.map((tx) => deserializePendingTx(tx, env));
```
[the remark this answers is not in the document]
elias-garcia: When you do `pendingTxParser(env).safeParse(tx)` you are already parsing a `SerializedPendingTx`, as they are being parsed. What's the point of parsing something that we already know its type? In this case we only want a function to convert between a `SerializedPendingTx` and a `PendingTx`, but a parser doesn't apply here. It can throw of course, in the same way a parser can throw!

You can do it in a functional way or in an imperative way... `throw catch` vs `if else`. Take a look at the last commit to see the functional implementation without using a parser. You functional guys are hard to fight! hahaha

src/utils/serializers.ts
elias-garcia: Not sure about this movement... Can you open an issue to discuss this further in the next weeks pls?

src/adapters/ethereum.ts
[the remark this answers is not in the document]
amonsosanz: Ok.

src/utils/serializers.ts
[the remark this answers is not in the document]
amonsosanz: Fair enough

src/utils/serializers.ts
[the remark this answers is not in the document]
amonsosanz: `deserializeBridgeId`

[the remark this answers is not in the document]
amonsosanz: I meant why not do this as we always do:

```diff
diff --git a/src/utils/serializers.ts b/src/utils/serializers.ts
index 7e7c0fd..f5f53db 100644
--- a/src/utils/serializers.ts
+++ b/src/utils/serializers.ts
@@ -97,12 +97,6 @@ const serializedPendingTxParser = StrictSchema<SerializedPendingTx>()(
   z.union([serializedPendingTxDepositParser, serializedPendingTxClaimParser])
 );

-const deserializeBridgeId = (
-  value: unknown
-): z.SafeParseReturnType<SerializedBridgeId, BridgeId> => {
-  return bridgeIdParser.safeParse(value);
-};
-
 const serializeBridgeId = (parsedBridgeId: BridgeId): SerializedBridgeId => {
   const { networkId, depositCount } = parsedBridgeId;
   return `${networkId}-${depositCount}`;
@@ -199,8 +193,8 @@ const deserializePendingTx = (
 };

 export {
+  bridgeIdParser,
   serializedPendingTxParser,
-  deserializeBridgeId,
   serializeBridgeId,
   serializePendingTx,
   deserializePendingTx,
diff --git a/src/views/bridge-details/bridge-details.view.tsx b/src/views/bridge-details/bridge-details.view.tsx
index 2bf209d..94561f6 100644
--- a/src/views/bridge-details/bridge-details.view.tsx
+++ b/src/views/bridge-details/bridge-details.view.tsx
@@ -23,7 +23,7 @@ import { AsyncTask, isMetamaskUserRejectedRequestError } from "src/utils/types";
 import { getBridgeStatus, getChainName, getCurrencySymbol } from "src/utils/labels";
 import { formatTokenAmount, formatFiatAmount, multiplyAmounts } from "src/utils/amounts";
 import { calculateTransactionResponseFee } from "src/utils/fees";
-import { deserializeBridgeId } from "src/utils/serializers";
+import { bridgeIdParser } from "src/utils/serializers";
 import { Bridge } from "src/domain";
 import routes from "src/routes";
 import { FIAT_DISPLAY_PRECISION, getEtherToken } from "src/constants";
@@ -123,7 +123,7 @@ const BridgeDetails: FC = () => {

   useEffect(() => {
     if (env && account.status === "successful") {
-      const parsedBridgeId = deserializeBridgeId(bridgeId);
+      const parsedBridgeId = bridgeIdParser.safeParse(bridgeId);
       if (parsedBridgeId.success) {
         const { depositCount, networkId } = parsedBridgeId.data;
         void getBridge({
```

We only create fractions that wrap parsers when we need some extra param in the parser. Otherwise, we just export and use the raw parser.

It's not very relevant. Do as you prefer.

src/utils/serializers.ts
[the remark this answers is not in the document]
amonsosanz: I still see te numbers, did you push the change?

src/adapters/storage.ts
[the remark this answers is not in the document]
amonsosanz: Ok

src/adapters/storage.ts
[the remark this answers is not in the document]
amonsosanz: > When you do `pendingTxParser(env).safeParse(tx)` you are already parsing a `SerializedPendingTx`, as they are being parsed in this line: `const serializedPendingTxs = getSerializedPendingTxs();`.

Absolutely! That part was beautifully implemented. We were safely parsing what we got from the local storage, i.e the serializable entity.

> What's the purpose of parsing something that we already know its type?

You're right in that we don't *have to* use a parser in this case since we can do this check manually. I proposed using a parser just because it was a shorter and simpler implementation.

> In this case we only want a function to convert between a `SerializedPendingTx` and a `PendingTx`, but a parser doesn't apply here. It can throw of course, in the same way a parser can throw!

Well, a parser would throw only if you use `parser.parse()`. If you use `parser.safeParse()` it would return an object with success true/false and the data or the error as your last implementation does. The new implementation is ok! You have removed the throw and now return the success true/false object. Basically, we're doing manually what the parser does internally. More verbose, but perfectly safe.

> You can do it in a functional way or in an imperative way... `throw catch` vs `if else`. Take a look at the last commit to see the functional implementation without needing to use a parser twice. You functional guys are hard to fight! hahaha

I don't see how this relates to functional vs. imperative. My comment was about safe vs. unsafe.

See the initial implementation behavior when the local storage contained the `from: 3` I mentioned:

image

It was breaking the app because the function `deserializePendingTx` was throwing.

When it comes to parsing the serialized TX of the localStorage we can do it:

* With 2 parsers (what I proposed)
* With 1 parser and a function that manually does the job of the second parser (current implementation)
* With 1 parser with the signature `StrictSchema<SerializedPendingTx, PendingTx>` which first takes care of parsing the primitives we get from the local storage and second takes care of inflating the `networkId`s to the actual `Chain`. And any of the steps fail, it throws with a parsing error.

src/adapters/storage.ts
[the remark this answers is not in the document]
amonsosanz: Sorry, I probably didn't make myself clear in the previous comment.

If you look at the rest of the functions in the `storage.ts` module you will see that all they return the new version of the local storage after doing the required manipulation (addition/removal of data).

I'd like us to honor this consistency with the new functions added in this PR.

If the consumer of these functions does not need the new version of the data present in the local storage, can just ignore it.

src/utils/serializers.ts
amonsosanz: Sure! [Here it's](https://github.com/0xPolygonHermez/zkevm-bridge-ui/issues/176).

src/contexts/bridge.context.tsx
amonsosanz: Ah, I see! I said about the TS shut up because I thought it was there to force the function to return `void` instead of `void[]`!

It's perfect now!

src/views/activity/activity.view.tsx
amonsosanz: No, it's not related to the number of bridges. The problem is that the state we use now is not fresh and therefore if there were already disabled bridges it does not work as expected. Look:


https://user-images.githubusercontent.com/15896177/194038080-81067a17-f53f-4aec-89fe-df0d10897389.mp4

src/views/activity/activity.view.tsx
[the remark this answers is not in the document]
amonsosanz: I still don't get this. To my understanding, whenever a bridge becomes available in the API we need to remove its shadow copy in the local storage. It should never show repeated and the API should always be the source of truth. Maybe we can jump on a call and show me what use case is covered by this `filteredApiBridges`.

src/views/bridge-details/bridge-details.view.tsx
elias-garcia: LOL right! What an amazing ternarny I did 🤣 Fixed!

src/contexts/bridge.context.tsx
```diff
@@ -568,6 +593,115 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
...
+              : undefined;
+          const fiatAmount =
+            tokenPrice &&
+            multiplyAmounts(
+              {
+                value: tokenPrice,
+                precision: FIAT_DISPLAY_PRECISION,
+              },
+              {
+                value: BigNumber.from(tx.amount),
```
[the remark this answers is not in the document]
elias-garcia: Right, fixed as well

src/views/activity/activity.view.tsx
elias-garcia: Well, in fact the behavior for this should be different. While you are waiting for tx confirmation you shouldn't be able to finalise more bridges. I'll revert it by the moment but I'll create an issue for improving this.

src/utils/serializers.ts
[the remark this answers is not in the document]
elias-garcia: We can address this in the issue you opened about reviewing this serializers module as well

src/utils/serializers.ts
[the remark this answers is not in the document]
elias-garcia: Mmmm yep, I see the right values now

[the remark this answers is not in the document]
elias-garcia: Done!

src/utils/serializers.ts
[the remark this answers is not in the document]
amonsosanz: I'd not say they are very different concepts, but anyway. We have now an [issue](https://github.com/0xPolygonHermez/zkevm-bridge-ui/issues/176) to discuss all this.

src/adapters/storage.ts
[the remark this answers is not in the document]
amonsosanz: Since we're now in a hurry I leave this comment open for future consideration.

src/views/activity/activity.view.tsx
```diff
@@ -48,19 +49,26 @@ const Activity: FC = () => {
...
+          }
           openSnackbar({
             type: "success-msg",
-            text: "Transaction successfully submitted.\nThe list will be updated once it is processed.",
+            text: "Transaction successfully submitted.",
           });
         })
         .catch((error) => {
-          setDisabledBridges((current) => current.filter((id) => id !== bridge.id));
+          setDisabledBridges((current) => [...current, bridge.id]);
```
amonsosanz, line 71: Why add instead of remove?

src/views/activity/activity.view.tsx
```diff
@@ -48,19 +49,26 @@ const Activity: FC = () => {
...
       claim({
         bridge,
       })
         .then(() => {
+          if (
+            bridgeList.status === "successful" ||
+            bridgeList.status === "reloading" ||
+            bridgeList.status === "loading-more-items"
+          ) {
+            processFetchBridgesSuccess(bridgeList.data);
```
amonsosanz, line 63: Why is this now required?

src/views/activity/activity.view.tsx
```diff
@@ -202,7 +224,11 @@ const Activity: FC = () => {
   }, [connectedProvider?.chainId]);
 
   const EmptyMessage = () => (
-    <Card className={classes.emptyMessage}>Bridge activity will be shown here</Card>
+    <Card className={classes.emptyMessage}>
+      {displayAll
+        ? "Bridge activity will be shown here"
+        : "There are no pending bridges at the moment"}
```
amonsosanz, line 230: Nice fix!

src/views/activity/activity.view.tsx
amonsosanz: It's required cause we need to add the new TX from the local storage.

src/views/activity/activity.view.tsx
amonsosanz: Fixed

amonsosanz (approved):

PR #186, opened 2022-10-06
Add Network box

### LINKED ISSUE #183: Display network info
In the home page of the bridge, we need to display all the details of the network:

- L2 RPC:
  - URL
  - ChainID
  - Currency symbol
- L2 Explorer URL.
- L1 SC link (Etherscan link).
- A button to directly add the network to MetaMask.

adrifdez:
Closes #183 

### What does this PR does?

Add a box at the bottom of the home with information about the networks that we use

## Checklist

review them:


src/views/home/home.view.tsx
```diff
@@ -34,6 +35,10 @@ const Home = (): JSX.Element => {
     setFormData(undefined);
   };
 
+  const onChangeNetwork = (chain: Chain) => {
+    void changeNetwork(chain);
```
amonsosanz, line 39: We should catch this error and report it when `isMetamaskUserRejectedRequestError` evaluates as `false`. And we should do it using `callIfMounted`:

```typescript
changeNetwork(chain).catch((error) => {
  callIfMounted(() => {
    if (isMetamaskUserRejectedRequestError(error) === false) {
      notifyError(error);
    }
  });
});
```

src/views/shared/network-box/network-box.view.tsx
```diff
@@ -0,0 +1,77 @@
...
+                rel="noreferrer"
+                target="_blank"
+                className={classes.link}
+              >
+                {env.chains[0].poeContractAddress}
+              </a>
+            </Typography>
+          </li>
+        </ul>
+        <button
```
amonsosanz, line 65: Shouldn't we disable this button once the network is already added and selected? Otherwise, you click a button that does nothing...

adrifdez: How do you know that the user has that network?

src/views/home/home.view.tsx
adrifdez: I don't think so because in this case the action is `Add to MetaMask` and if the user cancels it, I think that the experience has stopped and the user knows it

adrifdez: Sorry I was thinking in the other flow

src/views/shared/network-box/network-box.view.tsx
amonsosanz: Done

amonsosanz (approved):

PR #187, opened 2022-10-07
Display network name everywhere

### LINKED ISSUE #185: Display network name everywhere
The aim of this issue is to show a network selector button that will always show (and be synchronized) with the network selected in Metamask. This new button will be accessible on every screen of the dapp (but the login page) and will serve two purposes:

* Reliably display on all screens which network the dapp is currently connected to.
* Allow the user to change the connected network from the dapp itself, without having to do it in Metamask.

The new *Network switch* button will be positioned on every screen in the top right corner of the dapp.

This space is already available on all the screens but the Home screen, which currently shows the *Activity* button, so the Home screen will require some layout changes:

* The *Activity* button will be moved to the left side, next to the *Settings* button.
* Both the *Activity* and *Settings* buttons will become icon buttons (with no text).

Another significant change is that the *From* network selector in the form will no longer be a selector. Instead, it becomes a mere informative label of the network (exactly as *To* behaves) which will let the user know that the transaction (deposit) will start *from* the selected network.

amonsosanz:
Closes #185 

### What does this PR does?

This PR adds a network selector button that always shows (and is synchronized with) the network selected in Metamask. This button is accessible on every screen of the dapp (but the login page) and serves two purposes:

* Reliably displays on all screens which network the dapp is currently connected to.
* Allows the user to change the connected network from the dapp itself, without having to do it in Metamask.

One significant change is that the *From* network selector in the form is no longer a selector. Instead, it becomes a mere informative label of the network (exactly as *To* behaves) which lets the user know that the transaction (deposit) will start *from* the selected network.


src/views/home/components/header/header.styles.ts
```diff
@@ -10,17 +10,26 @@ const useHeaderStyles = createUseStyles((theme: Theme) => ({
...
-    alignItems: "center",
-    textAlign: "center",
-    flexDirection: "column",
+    justifyContent: "left",
+    gap: theme.spacing(0.75),
   },
-  appName: {
-    marginTop: theme.spacing(0.5),
-    color: theme.palette.black,
+  centerBlock: {
```
adrifdez, line 19: If the other blocks have specific `flex: 1`, you can leave this one just with `display: flex` or remove it and assume and small pixel in the bottom

src/views/shared/header/header.styles.ts
```diff
@@ -16,17 +16,34 @@ const useHeaderStyles = createUseStyles((theme: Theme) => ({
...
+  bottomRow: {
+    marginTop: theme.spacing(1.25),
+  },
+  leftBlock: {
+    flex: 1,
+    display: "flex",
+    justifyContent: "left",
+    gap: theme.spacing(0.75),
+  },
+  centerBlock: {
```
adrifdez, line 34: You can remove it

src/views/home/components/header/header.styles.ts
```diff
@@ -10,17 +10,26 @@ const useHeaderStyles = createUseStyles((theme: Theme) => ({
...
+    flex: 1,
+    display: "flex",
+    justifyContent: "center",
+    gap: theme.spacing(0.75),
```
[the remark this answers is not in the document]
amonsosanz: Yep, it was deliberate (not a leftover), I thought it was a good idea to prepare the block for further children. There are several unused gaps in this PR. Do you think I should remove them?

index.html
```diff
@@ -15,7 +15,7 @@
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
-    <div id="fullscreen-modal"></div>
+    <div id="fullscreen-modal" style="z-index: 1"></div>
```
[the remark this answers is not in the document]
adrifdez: Yes for that the modal should be the highest number and leave to the others to use smaller numbers

src/views/home/components/header/header.styles.ts
amonsosanz: Yep, I consciously added flex 1 to the 3 blocks to be explicit about them having exactly the same size.

src/views/shared/header/header.styles.ts
amonsosanz: Yes, I know it has no effect but I prefer the explicitness.

src/views/home/components/header/header.styles.ts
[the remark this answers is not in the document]
amonsosanz: As agreed, now block rules are abstracted.

adrifdez (approved):

PR #193, opened 2022-10-10
Improve the login flow

### LINKED ISSUE #144: Improve the login flow
Currently we are asking the user to connect again to the dapp even if he already has MetaMask connected to it. We need to investigate if we can avoid this login screen when the user is already logged in.

amonsosanz:
Closes #144

### What does this PR does?

This issue improves the application login process, automatically logging in (connecting the provider) when the MetaMask account is already connected to the dapp.

It also removes the logout button (which no longer makes sense, since it only disconnected the provider), and therefore, hides the Settings section when there are no data to show there.

### DIFF
--- a/src/adapters/ethereum.ts
+++ b/src/adapters/ethereum.ts
@@ -12,6 +12,15 @@ import { Token, Chain, TxStatus } from "src/domain";
 
 const ethereumAccountsParser = StrictSchema<string[]>()(z.array(z.string()));
 
+const silentlyGetConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
+  if (provider.provider.request) {
+    return provider.provider
+      .request({ method: "eth_accounts" })
+      .then((accounts) => ethereumAccountsParser.parse(accounts));
+  }
+  throw Error("No request method is available from the provider to get the Ethereum accounts");
+};
+
 const getConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
   return provider
     .send("eth_requestAccounts", [])
@@ -217,6 +226,7 @@ function hasTxBeenReverted(txReceipt: TransactionReceipt): boolean {
 
 export {
   ethereumAccountsParser,
+  silentlyGetConnectedAccounts,
   getConnectedAccounts,
   isPermitSupported,
   approve,
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,7 +1,7 @@
 import { ethers } from "ethers";
 import { JsonRpcProvider } from "@ethersproject/providers";
 
-import { Chain, Currency, EthereumChain, Token, ZkEVMChain } from "src/domain";
+import { Env, Chain, Currency, EthereumChain, Token, ZkEVMChain } from "src/domain";
 import { ReactComponent as EthChainIcon } from "src/assets/icons/chains/ethereum.svg";
 import { ReactComponent as PolygonZkEVMChainIcon } from "src/assets/icons/chains/polygon-zkevm.svg";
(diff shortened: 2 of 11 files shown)


src/adapters/ethereum.ts
```diff
@@ -12,6 +12,15 @@ import { Token, Chain, TxStatus } from "src/domain";
 
 const ethereumAccountsParser = StrictSchema<string[]>()(z.array(z.string()));
 
+const silentlyGetConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
+  if (provider.provider.request) {
```
elias-garcia, line 16: I would throw the error first:

```
if (!provider.provider.request) {
  throw Error("No request method is available from the provider to get the Ethereum accounts");
}

return provider.provider
      .request({ method: "eth_accounts" })
      .then((accounts) => ethereumAccountsParser.parse(accounts));
```

src/constants.ts
```diff
@@ -140,3 +148,7 @@ export const getUsdcToken = ({
   logoURI:
     "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/logo.png",
 });
+
+export const areSettingsVisible = (env: Env): boolean => {
```
elias-garcia, line 152: I would not create this function here just to return `env.fiatExchangeRates.areEnabled`.

amonsosanz: I did it this way to centralize the Decision Point that determines if the settings will be active since this condition may change in the future and be needed in more places

src/adapters/ethereum.ts
amonsosanz: Ok.

src/constants.ts
elias-garcia: Taking into account that this is a temporary solution I would just remove this function

src/views/shared/private-route/private-route.view.tsx
```diff
@@ -1,20 +1,63 @@
...
+    }
+    case "login": {
+      return (
+        <Navigate to={routes.login.path} replace state={{ redirectUrl: `${pathname}${search}` }} />
+      );
+    }
+    default: {
+      return null;
+    }
+  }
```
amonsosanz, line 60: Similar to the [comment in the login view](https://github.com/0xPolygonHermez/zkevm-bridge-ui/pull/193/files#r995531151), here we're overcomplicating things by having an extra state `nextRoute` and a `useEffect` instantiating the provider (this should ideally only happen in the provider context), and calling `silentlyGetMetaMaskConnectedAccounts` and `connectMetaMaskProvider`.

Before, using the `isProviderConnecting` flag this was as simple as:

```
  if (isProviderConnecting) {
    return null;
  } else if (!connectedProvider) {
    return (
      <Navigate to={routes.login.path} replace state={{ redirectUrl: `${pathname}${search}` }} />
    );
  } else {
    return <>{children}</>;
  }
```

I'm not convinced by all these changes to be honest.

While I agree with your initial comment regarding the `isProviderConnecting` flag in the provider being a bit hard to follow and bug-prone, it has the benefit of encapsulating all the provider-related logic (instantiating, connecting, etc) in the Provider module. This alternative is spreading the logic to the views and making things more complex outside the Provider module.

I think we need to keep thinking about this to find a better solution.

Random thought, what if, instead of adding the `isProviderConnecting` flag, we convert the provider to an async task (or a similar interface) so we can have it in "loading"/"logging-in" and the views return these nulls while this is its state?

Maybe we can have a meeting to keep discussing all this. Luckily, we're not in a hurry for this one.

amonsosanz:
This PR is closed in favor of #209.

PR #203, opened 2022-10-13
Show the network info box in the login screen

### LINKED ISSUE #202: Show the network box in the login screen
The purpose of this issue is to display the network box on the login screen while we keep it as it's on the Home screen.

amonsosanz:
Closes #202 

### What does this PR does?

This PR adds the network info box to the login screen.

### DIFF
--- a/src/contexts/providers.context.tsx
+++ b/src/contexts/providers.context.tsx
@@ -186,7 +186,7 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
           const { chainId } = await connectedProvider.getNetwork();
 
           if (chainId !== chain.chainId) {
-            return Promise.reject(new Error("Could not switch the network"));
+            throw "wrong-network";
           }
         })
         .catch((error) => {
@@ -217,7 +217,7 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
           const { chainId } = await connectedProvider.getNetwork();
 
           if (chainId !== chain.chainId) {
-            return Promise.reject(new Error("Could not switch the network"));
+            throw "wrong-network";
           }
         })
         .catch((error) => {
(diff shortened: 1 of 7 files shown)


src/views/shared/network-box/network-box.view.tsx
```diff
@@ -1,76 +1,113 @@
...
+    addNetwork(polygonZkEVMChain)
+      .then(() => {
+        setIsAddNetworkButtonDisabled(false);
+        callIfMounted(() => {
+          openSnackbar(successMsg);
+        });
+      })
+      .catch((error) => {
+        callIfMounted(() => {
+          setIsAddNetworkButtonDisabled(false);
```
elias-garcia, line 48: I would move this line to a `finally` to avoid repeating it.

src/contexts/providers.context.tsx
```diff
@@ -186,7 +190,7 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
           const { chainId } = await connectedProvider.getNetwork();
 
           if (chainId !== chain.chainId) {
-            return Promise.reject(new Error("Could not switch the network"));
+            throw "wrong-network";
```
elias-garcia, line 193: I would throw an Error to have more info in the object `message`, `stacktrace` etc:

```
throw new Error("wrong-network");
```

src/views/shared/network-box/network-box.view.tsx
amonsosanz: Yep, done.

src/contexts/providers.context.tsx
amonsosanz: As agreed, since throwing string errors is a common practice in the app we will tackle this in a dedicated task.

elias-garcia (approved):

PR #206, opened 2022-10-14
Show an error when there's not enough balance to pay the TX fee

### LINKED ISSUE #200: Fix out-of-bounds error estimating fee when left balance is less than fee
This issue is to show an error when there's not enough balance to pay the TX fee:

fee-balance-error

amonsosanz:
Closes #200

### What does this PR does?

This PR disables the Bridge button in the Confirmation screen and shows an error when there is not enough balance in the account to pay for the TX fee.

fee-balance-error


amonsosanz:
> What I would avoid here is to display a negative number in the amount. Instead, together wit the error I would display user account balance as the amount and the fee as we are already doing.

So you're proposing that we should not make the TX amount dynamic and always show the user input to avoid negative values? hmm... I don't like it. I think it is very confusing to see an amount that is not the one that you are really going to transfer. Let's discuss this in a meeting.

amonsosanz:
> > What I would avoid here is to display a negative number in the amount. Instead, together wit the error I would display user account balance as the amount and the fee as we are already doing.
> 
> So you're proposing that we should not make the TX amount dynamic and always show the user input to avoid negative values? hmm... I don't like it. I think it is very confusing to see an amount that is not the one that you are really going to transfer. Let's discuss this in a meeting.

As agreed in the meeting, we will show 0 when the TX amount is negative and will show the extra required ETH in the error string:

image

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -262,10 +263,20 @@ const BridgeConfirmation: FC = () => {
...
+                      newMaxPossibleAmountConsideringFee.isNegative()
+                        ? "0"
+                        : formatTokenAmount(newMaxPossibleAmountConsideringFee, token);
+
                     const hasOnScreenAmountChanged =
-                      oldMaxPossibleAmountConsideringFee !== undefined &&
-                      formatTokenAmount(oldMaxPossibleAmountConsideringFee, token) !==
-                        formatTokenAmount(newMaxPossibleAmountConsideringFee, token);
+                      formattedOldMaxPossibleAmountConsideringFee !==
+                      formattedNewMaxPossibleAmountConsideringFee;
```
elias-garcia, line 279: Why do we need to format both `oldMaxPossibleAmountConsideringFee` and `newMaxPossibleAmountConsideringFee` to check if the amount has changed? We can just use the `.eq` method in `BigNumber` to check if they are equal or not.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -75,6 +75,7 @@ const BridgeConfirmation: FC = () => {
   const [estimatedFee, setEstimatedFee] = useState<AsyncTask<BigNumber, string>>({
     status: "pending",
   });
+  const feePollingInterval = useRef<NodeJS.Timer>();
```
elias-garcia, line 78: We shouldn't make any reference to `NodeJS` types in the browser. This should be `useRef<number>`, and then later you should use `window.setInterval` instead of just `setInterval`.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: There are 2 reasons for this:
* When the value is negative we use `"0"` as the formatted value, so two different negative values will format the same.
* The formatted value is using less precision than the actual number so two different positive values can format the same.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: Good catch! I didn't know that typing this as `NodeJS.Timer` could imply runtime errors since it's a number in the browser. Fixed.

elias-garcia (approved):

PR #209, opened 2022-10-14
Improve the login flow

### LINKED ISSUE #144: Improve the login flow

amonsosanz:
Closes #144

### What does this PR does?

This issue improves the application login process, automatically logging in (connecting the provider) when the MetaMask account is already connected to the dapp.

It also removes the logout button (which no longer makes sense, since it only disconnected the provider), and therefore, hides the Settings section when there are no data to show there.

### DIFF
--- a/src/adapters/ethereum.ts
+++ b/src/adapters/ethereum.ts
@@ -13,12 +13,12 @@ import { Token, Chain, TxStatus } from "src/domain";
 const ethereumAccountsParser = StrictSchema<string[]>()(z.array(z.string()));
 
 const silentlyGetConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
-  if (provider.provider.request) {
-    return provider.provider
-      .request({ method: "eth_accounts" })
-      .then((accounts) => ethereumAccountsParser.parse(accounts));
+  if (!provider.provider.request) {
+    throw Error("No request method is available from the provider to get the Ethereum accounts");
   }
-  throw Error("No request method is available from the provider to get the Ethereum accounts");
+  return provider.provider
+    .request({ method: "eth_accounts" })
+    .then((accounts) => ethereumAccountsParser.parse(accounts));
 };
 
 const getConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
--- a/src/adapters/ethereum.ts
+++ b/src/adapters/ethereum.ts
@@ -12,6 +12,15 @@ import { Token, Chain, TxStatus } from "src/domain";
 
 const ethereumAccountsParser = StrictSchema<string[]>()(z.array(z.string()));
 
+const silentlyGetConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
+  if (provider.provider.request) {
+    return provider.provider
+      .request({ method: "eth_accounts" })
+      .then((accounts) => ethereumAccountsParser.parse(accounts));
+  }
+  throw Error("No request method is available from the provider to get the Ethereum accounts");
+};
+
 const getConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
   return provider
     .send("eth_requestAccounts", [])
@@ -217,6 +226,7 @@ function hasTxBeenReverted(txReceipt: TransactionReceipt): boolean {
 
 export {
   ethereumAccountsParser,
+  silentlyGetConnectedAccounts,
   getConnectedAccounts,
   isPermitSupported,
   approve,
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,7 +1,7 @@
 import { ethers } from "ethers";
 import { JsonRpcProvider } from "@ethersproject/providers";
 
-import { Chain, Currency, EthereumChain, Token, ZkEVMChain } from "src/domain";
+import { Env, Chain, Currency, EthereumChain, Token, ZkEVMChain } from "src/domain";
 import { ReactComponent as EthChainIcon } from "src/assets/icons/chains/ethereum.svg";
 import { ReactComponent as PolygonZkEVMChainIcon } from "src/assets/icons/chains/polygon-zkevm.svg";
 import { ProofOfEfficiency__factory } from "src/types/contracts/proof-of-efficiency";
@@ -137,3 +137,7 @@ export const getUsdcToken = ({
   logoURI:
     "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/logo.png",
 });
+
+export const areSettingsVisible = (env: Env): boolean => {
+  return env.fiatExchangeRates.areEnabled;
+};
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -5,15 +5,7 @@ import { ReactComponent as EthChainIcon } from "src/assets/icons/chains/ethereum
 import { ReactComponent as PolygonZkEVMChainIcon } from "src/assets/icons/chains/polygon-zkevm.svg";
 import { ProofOfEfficiency__factory } from "src/types/contracts/proof-of-efficiency";
 import { ProviderError } from "src/adapters/error";
-import {
-  Env,
-  Chain,
-  Currency,
-  EthereumChain,
-  EthereumChainId,
-  Token,
-  ZkEVMChain,
-} from "src/domain";
+import { Chain, Currency, EthereumChain, EthereumChainId, Token, ZkEVMChain } from "src/domain";
 
 export const UNISWAP_V2_ROUTER_02_CONTRACT_ADDRESS = "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D";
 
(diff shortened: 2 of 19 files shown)


src/views/shared/private-route/private-route.view.tsx
```diff
@@ -8,13 +8,21 @@ const PrivateRoute: FC<PropsWithChildren> = ({ children }) => {
...
   const { pathname, search } = useLocation();
 
-  if (!connectedProvider) {
-    return (
-      <Navigate to={routes.login.path} replace state={{ redirectUrl: `${pathname}${search}` }} />
-    );
+  switch (connectedProvider.status) {
+    case "pending":
+    case "loading":
+    case "reloading": {
```
elias-garcia, line 14: What happens if we return `<>{children}</>` in this case? It would definitely look better than a reload when switching the provider.

src/contexts/providers.context.tsx
```diff
@@ -18,38 +18,55 @@ import {
...
 import routes from "src/routes";
-import { Chain, EthereumChainId, EthereumEvent, WalletName } from "src/domain";
+import { Env, Chain, EthereumChainId, EthereumEvent, WalletName } from "src/domain";
 
 interface ProvidersContext {
-  connectedProvider?: { provider: Web3Provider; chainId: number };
+  connectedProvider: AsyncTask<
+    {
+      provider: Web3Provider;
+      chainId: number;
```
elias-garcia, line 37: Probably it would make sense to add the `account` here as well instead of having it separately, as the `connectedProvider` that we are using is strictly connected to an `account`.

elias-garcia: I would create a separate interface for this in our domain btw.

src/contexts/providers.context.tsx
```diff
@@ -60,45 +77,79 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
...
+              ? `Switch your network to Ethereum Goerli testnet or ${env.chains[1].name} to continue`
+              : `Switch your network to Ethereum or ${env.chains[1].name} to continue`,
+          });
+        } else {
+          setAccount({ status: "successful", data: account });
+          setConnectedProvider({
+            status: "successful",
+            data: { provider: web3Provider, chainId: currentNetworkChainId },
+          });
+          return setAccount({ status: "successful", data: account });
```
elias-garcia, line 115: Yo are setting the account twice here and in L110.

src/contexts/providers.context.tsx
```diff
@@ -109,76 +160,50 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
             if (isMetaMaskResourceUnavailableError(error)) {
               return setAccount({
                 status: "failed",
-                error: `Please unlock ${WalletName.METAMASK} to continue`,
+                error: `Please connect to or unlock ${WalletName.METAMASK} to continue`,
```
elias-garcia, line 163: `Please unlock or connect to...` sounds better.

src/contexts/providers.context.tsx
```diff
@@ -259,15 +287,34 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
...
   useEffect(() => {
     if (account.status === "failed") {
       navigate(routes.login.path);
     }
   }, [account, navigate]);
 
   useEffect(() => {
-    const internalConnectedProvider: Record<string, unknown> | undefined =
-      connectedProvider?.provider.provider;
+    const externalConnectedProvider: Record<string, unknown> | undefined = isAsyncTaskDataAvailable(
```
elias-garcia, line 313: How is the type of this `externalConnectedProvider` variable `Record<string, unknown> | undefined`? Shouldn't this type be `ExternalProvider | undefined`?.

elias-garcia: I would just use `externalProvider` as the name as it's even more correct in terms of `ethers`.

elias-garcia (commented):
I see that the `onDisconnect` function in the providers context is still reloading the page:

```
const onDisconnect = () => {
  window.location.reload();
};
```

src/adapters/ethereum.ts
```diff
@@ -12,6 +12,15 @@ import { Token, Chain, TxStatus } from "src/domain";
 
 const ethereumAccountsParser = StrictSchema<string[]>()(z.array(z.string()));
 
+const silentlyGetConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
+  if (provider.provider.request) {
```
elias-garcia, line 16: As I proposed in the other PR I would throw first:

```
if (!provider.provider.request) {
  throw Error("No request method is available from the provider to get the Ethereum accounts");
}

return provider.provider
  .request({ method: "eth_accounts" })
  .then((accounts) => ethereumAccountsParser.parse(accounts));
```

src/constants.ts
```diff
@@ -140,3 +148,7 @@ export const getUsdcToken = ({
   logoURI:
     "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/logo.png",
 });
+
+export const areSettingsVisible = (env: Env): boolean => {
```
elias-garcia, line 152: As we've been talking I would move this function to a different module if we want to keep it.

src/views/shared/private-route/private-route.view.tsx
amonsosanz: It won't make any difference today since this AsyncTask is never set to `reloading`. Agreed though. What you propose is more correct and might be relevant tomorrow. Done.

src/contexts/providers.context.tsx
amonsosanz: Although it makes sense, this is a significant change. Are you sure about going it in this PR?

I can intuit changes in the behavior of certain reloads/renders in useEffects derived from the change of identity of the object in their dependencies.

src/contexts/providers.context.tsx
amonsosanz: Good catch! Fixed.

src/views/shared/private-route/private-route.view.tsx
elias-garcia: This is a good point indeed. Because I've seen that in some effects now we are using `isAsyncTaskDataAvailable` to calculate some things, but I think that we should just check the `successful` status but not the `reloading` one to avoid calculating things unnecessary when we use the `reloading` state at some point.

src/contexts/providers.context.tsx
elias-garcia: If those changes happens in the `useEffect` then it's perfect! That's how they should work, if we change the `connectedProvider` and hence the `account` data, things depending on the `account` should re-render.

I would try to do it in this PR as the last thing, and if we see that things get complicated I'll move it to another PR.

src/contexts/providers.context.tsx
```diff
@@ -60,45 +77,79 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
...
 
+  interface ConnectMetamaskProviderParams {
+    env: Env;
+    web3Provider: Web3Provider;
+    account: string;
+  }
+
+  const connectMetamaskProvider = useCallback(
+    async ({ env, web3Provider, account }: ConnectMetamaskProviderParams): Promise<void> => {
+      // ToDo: Why is this try/catch required in a promise?
```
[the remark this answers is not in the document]
amonsosanz: This comment was for me to check later and forgot about it 🤣.

I think it's better to have the error management centralized here.

Comment removed.

src/contexts/providers.context.tsx
amonsosanz: Yep, done.

src/contexts/providers.context.tsx
amonsosanz: > How is the type of this `externalConnectedProvider` variable `Record<string, unknown> | undefined`? Shouldn't this type be `ExternalProvider | undefined`?.

It's a record because the `ExternalProvider` type does not include the methods `on` and `removeListener`. More on this [here](https://github.com/ethers-io/ethers.js/issues/1396#issuecomment-806380431).

We declare it as a Record so we later can do `"on" in externalProvider && typeof externalProvider.on === "function"` and `"removeListener" in externalProvider && typeof externalProvider.removeListener === "function"` and call the methods.

We could override the type, but this approach is safer than overriding the type and cleaner than parsing the external provider instance, so I think it's ok as it's.

amonsosanz:
> I see that the `onDisconnect` function in the providers context is still reloading the page:
> 
> ```
> const onDisconnect = () => {
>   window.location.reload();
> };
> ```

Done! Completely forgot about this!

src/adapters/ethereum.ts
amonsosanz: Done!

src/constants.ts
amonsosanz: Done!

src/views/shared/private-route/private-route.view.tsx
amonsosanz: Done as agreed in our meeting. Now we use `connectedProvider.status === "successful"` in all the useEffect's.

src/contexts/providers.context.tsx
amonsosanz: Since this task implies several changes, I opened PR #212 pointing to this one so you can review these changes in isolation.

elias-garcia:
Looks very good! 👏🏼

elias-garcia (approved):

PR #212, opened 2022-10-18
Move the account to the connectedProvider AsyncTask

amonsosanz:
### What does this PR does?

This PR is complementary to #209 and focuses on including the `account` within the `connectedProvider` async task.


src/contexts/providers.context.tsx
```diff
@@ -297,7 +283,10 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
           if (account) {
             void connectMetamaskProvider({ env, web3Provider, account });
           } else {
-            setConnectedProvider({ status: "failed", error: "No connected accounts found" });
+            setConnectedProvider({
+              status: "failed",
+              error: "",
```
elias-garcia, line 288: I think that you forgot the error here: `"No connected accounts found"`

elias-garcia (commented):
Just one comment, I think that this is a huge improvement regarding how me manage the provider, it simplifies things a lot 🚀

src/contexts/providers.context.tsx
amonsosanz: It is a conscious change, otherwise, if we cannot do the automatic login we will show this error and I don't think we want that. We just want to let the user log in again right?

PR #217, opened 2022-10-20
Check if the current connected account is allowed to see the details of a bridge

elias-garcia:
### What does this PR does?

This PR adds a check to the `BridgeDetails` page to check the current user has participated in a Bridge, wether as a sender or a receiver.

## Checklist

review them:

### DIFF
--- a/src/contexts/providers.context.tsx
+++ b/src/contexts/providers.context.tsx
@@ -314,13 +314,21 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
       if (parsedAccounts.success && isAsyncTaskDataAvailable(connectedProvider)) {
         const account: string | undefined = parsedAccounts.data[0];
         if (account) {
-          setConnectedProvider({
-            status: "successful",
-            data: {
-              ...connectedProvider.data,
-              account: getChecksumAddress(account),
-            },
-          });
+          try {
+            setConnectedProvider({
+              status: "successful",
+              data: {
+                ...connectedProvider.data,
+                account: getChecksumAddress(account),
+              },
+            });
+          } catch (error) {
+            setConnectedProvider({
+              status: "failed",
+              error: "An error occurred connecting the provider",
+            });
+            notifyError(error);
+          }
         } else {
           setConnectedProvider({ status: "pending" });
         }
@@ -357,7 +365,7 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
         externalProvider.removeListener(EthereumEvent.DISCONNECT, onDisconnect);
       }
     };
-  }, [connectProvider, connectedProvider, navigate]);
+  }, [connectProvider, connectedProvider, navigate, notifyError]);
 
   const value = useMemo(
     () => ({
--- a/src/contexts/providers.context.tsx
+++ b/src/contexts/providers.context.tsx
@@ -35,6 +35,7 @@ import {
   WalletName,
   ConnectedProvider,
 } from "src/domain";
+import { getChecksumAddress } from "src/utils/addresses";
 
 interface ProvidersContext {
   connectedProvider: AsyncTask<ConnectedProvider, string>;
@@ -97,7 +98,11 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
         } else {
           setConnectedProvider({
             status: "successful",
-            data: { provider: web3Provider, chainId: currentNetworkChainId, account },
+            data: {
+              provider: web3Provider,
+              chainId: currentNetworkChainId,
+              account: getChecksumAddress(account),
+            },
           });
         }
       } catch (error) {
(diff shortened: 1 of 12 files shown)


src/domain/index.ts
```diff
@@ -126,6 +126,18 @@ interface BridgeCommonFields {
...
+  depositTxHash: BridgeCommonFields["depositTxHash"];
+  destinationAddress: BridgeCommonFields["destinationAddress"];
+  claimTxHash?: CompletedBridge["claimTxHash"];
+  from: BridgeCommonFields["from"];
+  to: BridgeCommonFields["to"];
+  token: BridgeCommonFields["token"];
+  amount: BridgeCommonFields["amount"];
+  fiatAmount: BridgeCommonFields["fiatAmount"];
+};
+
```
amonsosanz, line 140:
```typescript
export type PendingBridge = Pick<
  BridgeCommonFields,
  "depositTxHash" | "destinationAddress" | "from" | "to" | "token" | "amount" | "fiatAmount"
> & {
  status: "pending";
  claimTxHash?: CompletedBridge["claimTxHash"];
};
```

I'm not a fan of `Pick` but in this case, it has the advantage of being less verbose and clearly spotting the actual differences between the two types: the `status` and the `claimTxHash` coming from the `CompletedBridge` with the optional modifier.

What do you think?

src/views/bridge-details/bridge-details.view.tsx
```diff
@@ -133,7 +133,19 @@ const BridgeDetails: FC = () => {
...
+
+            if (!tx) {
+              throw new Error(`DepositTx not found: ${bridge.depositTxHash}`);
+            }
+
+            const allowedAccounts = [tx.from, bridge.destinationAddress];
+
+            if (!allowedAccounts.includes(connectedProvider.data.account)) {
+              return navigate(routes.activity.path);
+            }
```
amonsosanz, line 147: Why allow `tx.from` considering that the API of the bridge only returns the deposits belonging to the `destinationAddress`?

In other words, if I have the addresses `0xA` and `0xB` and I bridge some funds from `0xA` specifying `0xB` as `destinationAddress`, while the selected account in MetaMask is `0xA`, a request `/bridges/0xA` won't return this deposit.

If I need to change the account to `0xB` to see the deposit in the list, I think we should be consistent and hide the deposit from `0xA`.

The reasoning is based on the assumption that as a user (account) I should not see the details of a deposit which does not belong to me (i.e, is not in the list for the selected account).


```diff
diff --git a/src/views/bridge-details/bridge-details.view.tsx b/src/views/bridge-details/bridge-details.view.tsx
index aab33a2..363ff43 100644
--- a/src/views/bridge-details/bridge-details.view.tsx
+++ b/src/views/bridge-details/bridge-details.view.tsx
@@ -140,9 +140,7 @@ const BridgeDetails: FC = () => {
               throw new Error(`DepositTx not found: ${bridge.depositTxHash}`);
             }

-            const allowedAccounts = [tx.from, bridge.destinationAddress];
-
-            if (!allowedAccounts.includes(connectedProvider.data.account)) {
+            if (connectedProvider.data.account !== bridge.destinationAddress) {
               return navigate(routes.activity.path);
             }
```

elias-garcia: Mmmm you are right! I didn't remember that the API is returning only the deposits for the `destinationAddress`, so it doesn't make sense to take `tx.from` into account :'') Fixed.

src/domain/index.ts
elias-garcia: I'm a big fan of this kind of this! I didn't add it before because I know that you hate this beautiful utils 🤣

But now that we are going to add it... I don't like to reference `claimTxHash` from an interface below its declaration, so I'd do this:

```
export type PendingBridge = Pick<
  BridgeCommonFields,
  "depositTxHash" | "destinationAddress" | "from" | "to" | "token" | "amount" | "fiatAmount"
> & {
  status: "pending";
  claimTxHash?: string;
};

export type CompletedBridge = BridgeCommonFields & {
  status: "completed";
  claimTxHash: NonNullable<PendingBridge["claimTxHash"]>;
};
```

Now we are using the beautiful things of TS 😎

src/contexts/providers.context.tsx
```diff
@@ -307,7 +316,10 @@ const ProvidersProvider: FC<PropsWithChildren> = (props) => {
         if (account) {
           setConnectedProvider({
             status: "successful",
-            data: { ...connectedProvider.data, account },
+            data: {
+              ...connectedProvider.data,
+              account: getChecksumAddress(account),
```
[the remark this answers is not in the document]
elias-garcia: Right, i totally forgot to add it here! Added now.

src/domain/index.ts
amonsosanz: this `NonNullable<PendingBridge["claimTxHash"]>` hurts

src/views/bridge-details/bridge-details.view.tsx
```diff
@@ -134,6 +134,10 @@ const BridgeDetails: FC = () => {
           networkId,
         })
           .then((bridge) => {
+            if (bridge.destinationAddress !== connectedProvider.data.account) {
+              return navigate(routes.activity.path);
+            }
```
amonsosanz, line 139: I think this should be inside the `callIfMounted` block below. I guess we don't want to redirect the user when she has already left the page.

amonsosanz (approved):

PR #222, opened 2022-10-23
Fix ERR_CANCELED error and prevent extra request on activity and bridge details initial load

### LINKED ISSUE #219: Fix ERR_CANCELED error and prevent extra request on Activity load
The Activity screen of the Bridge implements a polling to always show the fresh status of the bridges to the user.

The app uses the Axios' [cancel token API](https://axios-http.com/docs/cancellation) to abort an in-progress polling request when the user requests a new page. After the new page is retrieved the polling is resumed.

Currently, the `CanceledError` of the cancel token API is not properly managed, and therefore when a polling request is canceled the UI shows the report error bar with the error:

```An unknown error has occurred: {"message":"canceled","name":"CanceledError","code":"ERR_CANCELED","status":null}```

image

Considering that the cancel token API is currently deprecated in favor of the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) to cancel requests, this issue is mainly focused on these tasks:

* Replacing the cancel token API with the AbortController.
* Using the Axios util `isCancel` to easily spot and ignore cancelation errors.
* Prevent extra requests to the API when the Activity page is initially loaded.
* Prevent extra requests to the API when the Bridge Details page is initially loaded.

### PR DESCRIPTION
**Pull request by amonsosanz**
Closes #219

### What does this PR does?

The Activity screen of the Bridge implements a polling to always show the fresh status of the bridges to the user.

The app uses the Axios' [cancel token API](https://axios-http.com/docs/cancellation) to abort an in-progress polling request when the user requests a new page. After the new page is retrieved the polling is resumed.

Currently, the `CanceledError` of the cancel token API is not properly managed, and therefore when a polling request is canceled the UI shows the report error bar with the error:

```An unknown error has occurred: {"message":"canceled","name":"CanceledError","code":"ERR_CANCELED","status":null}```

image

Considering that the cancel token API is currently deprecated in favor of the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) to cancel requests this PR tackles the following:

* Replacing the cancel token API with the AbortController.
* Using the Axios util `isCancel` to easily spot and ignore cancelation errors.
* Prevent extra requests to the API when the Activity page is initially loaded.
* Prevent extra requests to the API when the Bridge Details page is initially loaded.

### DIFF
--- a/src/views/bridge-details/bridge-details.view.tsx
+++ b/src/views/bridge-details/bridge-details.view.tsx
@@ -32,7 +32,7 @@ import routes from "src/routes";
 import { FIAT_DISPLAY_PRECISION, getEtherToken } from "src/constants";
 import useCallIfMounted from "src/hooks/use-call-if-mounted";
 import { useTokensContext } from "src/contexts/tokens.context";
-import { isAxiosCancelRequestError } from "src/adapters/bridge-api";
+import { isCancelRequestError } from "src/adapters/bridge-api";
 import { Bridge } from "src/domain";
 
 interface Fees {
@@ -151,7 +151,7 @@ const BridgeDetails: FC = () => {
             });
           })
           .catch((error) => {
-            if (!isAxiosCancelRequestError(error)) {
+            if (!isCancelRequestError(error)) {
               callIfMounted(() => {
                 notifyError(error);
                 setBridge({
--- a/src/views/bridge-details/bridge-details.view.tsx
+++ b/src/views/bridge-details/bridge-details.view.tsx
@@ -124,7 +124,7 @@ const BridgeDetails: FC = () => {
   }, [connectedProvider, bridge]);
 
   useEffect(() => {
-    if (env && connectedProvider.status === "successful") {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/views/bridge-details/bridge-details.view.tsx [hunk 1]
```diff
@@ -91,6 +92,8 @@ const BridgeDetails: FC = () => {
     status: bridge.status === "successful" ? bridge.data.status : undefined,
   });
 
+  const getBridgeAbortController = useRef<AbortController>(new AbortController());
```
elias-garcia: `getBridgeAbortController` -> `bridgeAbortController`

amonsosanz: Why? this is the `AbortController` of the function `getBridge` so it's the `getBridgeAbortController`

amonsosanz: We have renamed `getBridge` as `fetchBridge` and this `AbortController` as `fetchBridgeAbortController`.

elias-garcia (approved):

PR #223, opened 2022-10-24
Improve fee estimation

### LINKED ISSUE #215: Improve fee polling
Currently, we are polling for the fees using an interval, but this doesn't make a lot of sense taking into account that the fee will change only when a new block is mined. So, we need to listen to the [`"block"` ethers provider event](https://docs.ethers.io/v5/api/providers/provider/#Provider--events) instead and estimate the bridge fee only when this event is fired.

elias-garcia:
Closes #215.

### What does this PR does?
~This PR improves the fee polling, we now estimate the `bridge` fees when a new block occurs instead of each 10 seconds.~

This PR removes querying for the fee each X seconds and takes into account a `gasPrice` increase when estimating the bridge feee.

## Checklist

review them:

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -36,7 +36,9 @@ export const PAGE_SIZE = 25;
 
 export const PENDING_TX_TIMEOUT = 30 * 60 * 1000; // 30min in ms
 
-export const BRIDGE_CALL_GAS_INCREASE_PERCENTAGE = 20; // 20%
+export const BRIDGE_CALL_GAS_LIMIT_INCREASE_PERCENTAGE = 20; // 20%
+
+export const GAS_PRICE_INCREASE_PERCENTAGE = 50; //
 
 export const REPORT_ERROR_FORM_ENTRIES = {
   url: "entry.2056392454",
--- a/src/hooks/use-debounced-block.ts
+++ b/src/hooks/use-debounced-block.ts
@@ -1,61 +1,30 @@
-import { useCallback, useEffect, useState } from "react";
+import { useEffect, useState } from "react";
 
+import { useDebounce } from "src/hooks/use-debounce";
+import { AsyncTask, isAsyncTaskDataAvailable } from "src/utils/types";
 import { ConnectedProvider } from "src/domain";
-import { AsyncTask } from "src/utils/types";
-
-type DebouncedBlockCallback = (blockNumber: number) => void;
 
 const DEBOUNCE_TIME_IN_MS = 750;
 
 const useDebouncedBlock = (connectedProvider: AsyncTask<ConnectedProvider, string>) => {
-  const [listener, setListener] = useState<DebouncedBlockCallback>();
   const [blockNumber, setBlockNumber] = useState<number>();
 
-  const addBlockListener = useCallback(
-    (cb: DebouncedBlockCallback) => {
-      if (!listener) {
-        setListener(() => cb);
-      }
-    },
-    [listener]
-  );
-
   useEffect(() => {
-    const timeoutId = setTimeout(() => {
-      if (listener && blockNumber) {
-        listener(blockNumber);
-      }
-    }, DEBOUNCE_TIME_IN_MS);
-
-    return () => {
-      clearTimeout(timeoutId);
-    };
-  }, [blockNumber, listener]);
-
-  useEffect(() => {
-    if (listener) {
-      const onBlock = (newBlockNumber: number) => {
-        if (!blockNumber || newBlockNumber > blockNumber) {
-          setBlockNumber(newBlockNumber);
-        }
-      };
-
-      if (connectedProvider.status === "successful") {
-        void connectedProvider.data.provider.getBlockNumber().then((currentBlockNumber) => {
-          onBlock(currentBlockNumber);
-          connectedProvider.data.provider.on("block", onBlock);
+    if (isAsyncTaskDataAvailable(connectedProvider)) {
+      void connectedProvider.data.provider
+        .getBlockNumber()
+        .then(setBlockNumber)
+        .then(() => {
+          connectedProvider.data.provider.on("block", setBlockNumber);
         });
-      }
 
       return () => {
-        if (connectedProvider.status === "successful") {
-          connectedProvider.data.provider.off("block", onBlock);
-        }
+        connectedProvider.data.provider.off("block", setBlockNumber);
       };
     }
-  }, [blockNumber, connectedProvider, listener]);
+  }, [connectedProvider]);
 
-  return addBlockListener;
+  return useDebounce(blockNumber, DEBOUNCE_TIME_IN_MS);
 };
 
 export default useDebouncedBlock;
--- a/src/hooks/use-debounced-block.ts
+++ b/src/hooks/use-debounced-block.ts
@@ -1,0 +1,61 @@
+import { useCallback, useEffect, useState } from "react";
+
+import { ConnectedProvider } from "src/domain";
+import { AsyncTask } from "src/utils/types";
+
+type DebouncedBlockCallback = (blockNumber: number) => void;
+
+const DEBOUNCE_TIME_IN_MS = 750;
+
+const useDebouncedBlock = (connectedProvider: AsyncTask<ConnectedProvider, string>) => {
+  const [listener, setListener] = useState<DebouncedBlockCallback>();
+  const [blockNumber, setBlockNumber] = useState<number>();
+
+  const addBlockListener = useCallback(
+    (cb: DebouncedBlockCallback) => {
+      if (!listener) {
+        setListener(() => cb);
+      }
+    },
+    [listener]
+  );
+
+  useEffect(() => {
+    const timeoutId = setTimeout(() => {
+      if (listener && blockNumber) {
+        listener(blockNumber);
+      }
+    }, DEBOUNCE_TIME_IN_MS);
+
+    return () => {
+      clearTimeout(timeoutId);
+    };
+  }, [blockNumber, listener]);
+
+  useEffect(() => {
+    if (listener) {
+      const onBlock = (newBlockNumber: number) => {
+        if (!blockNumber || newBlockNumber > blockNumber) {
+          setBlockNumber(newBlockNumber);
+        }
+      };
+
+      if (connectedProvider.status === "successful") {
+        void connectedProvider.data.provider.getBlockNumber().then((currentBlockNumber) => {
+          onBlock(currentBlockNumber);
+          connectedProvider.data.provider.on("block", onBlock);
+        });
+      }
+
+      return () => {
+        if (connectedProvider.status === "successful") {
+          connectedProvider.data.provider.off("block", onBlock);
+        }
+      };
+    }
+  }, [blockNumber, connectedProvider, listener]);
+
+  return addBlockListener;
+};
+
+export default useDebouncedBlock;
--- a/src/views/bridge-confirmation/bridge-confirmation.view.tsx
+++ b/src/views/bridge-confirmation/bridge-confirmation.view.tsx
@@ -39,6 +39,7 @@ import ApprovalInfo from "src/views/bridge-confirmation/components/approval-info
 import { Gas } from "src/domain";
 
 const FADE_DURATION_IN_SECONDS = 0.5;
+const FADE_DURATION_IN_MS = FADE_DURATION_IN_SECONDS * 1000;
 
 const BridgeConfirmation: FC = () => {
   const callIfMounted = useCallIfMounted();
@@ -68,6 +69,7 @@ const BridgeConfirmation: FC = () => {
   const [estimatedGas, setEstimatedGas] = useState<AsyncTask<Gas, string>>({
     status: "pending",
   });
+  const [fadeClass, setFadeClass] = useState<string>();
   const currencySymbol = getCurrencySymbol(getCurrency());
 
   useEffect(() => {
@@ -111,7 +113,11 @@ const BridgeConfirmation: FC = () => {
               setMaxAmountIncludingFee(amount);
             }
 
-            setEstimatedGas({ status: "successful", data: gas });
+            setFadeClass(classes.fadeOut);
+            setTimeout(() => {
+              setEstimatedGas({ status: "successful", data: gas });
+              setFadeClass(classes.fadeIn);
+            }, FADE_DURATION_IN_MS);
           })
           .catch((error) => {
             if (isEthersInsufficientFundsError(error)) {
@@ -137,6 +143,7 @@ const BridgeConfirmation: FC = () => {
     }
   }, [
     callIfMounted,
+    classes,
     connectedProvider,
     estimateBridgeGas,
     estimatedGas,
@@ -458,7 +465,7 @@ const BridgeConfirmation: FC = () => {
         </div>
         <div className={classes.feeBlock}>
           <Typography type="body2">Estimated gas fee</Typography>
-          <div className={classes.fee}>
+          <div className={`${classes.fee} ${fadeClass || ""}`}>
             <Icon url={ETH_TOKEN_LOGO_URI} size={20} />
             <Typography type="body1">{feeString}</Typography>
           </div>
--- a/src/views/bridge-confirmation/bridge-confirmation.view.tsx
+++ b/src/views/bridge-confirmation/bridge-confirmation.view.tsx
@@ -37,16 +37,10 @@ import BridgeButton from "src/views/bridge-confirmation/components/bridge-button
 import useBridgeConfirmationStyles from "src/views/bridge-confirmation/bridge-confirmation.styles";
 import ApprovalInfo from "src/views/bridge-confirmation/components/approval-info/approval-info.view";
 import { Gas } from "src/domain";
-import useDebouncedBlock from "src/hooks/use-debounced-block";
-
-const FADE_DURATION_IN_SECONDS = 0.5;
-const FADE_DURATION_IN_MS = FADE_DURATION_IN_SECONDS * 1000;
 
 const BridgeConfirmation: FC = () => {
   const callIfMounted = useCallIfMounted();
-  const classes = useBridgeConfirmationStyles({
-    fadeDurationInSeconds: FADE_DURATION_IN_SECONDS,
-  });
+  const classes = useBridgeConfirmationStyles();
   const navigate = useNavigate();
   const env = useEnvContext();
   const { notifyError } = useErrorContext();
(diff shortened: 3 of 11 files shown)


src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -77,102 +69,55 @@ const BridgeConfirmation: FC = () => {
...
-                    data: newGas,
-                  });
-                  setIsFadeVisible(true);
-                }, msTimeout);
+            if (isTokenEther) {
+              const amountIncludingFee = amount.add(newFee);
+              const tokenBalanceRemainder = amountIncludingFee.sub(tokenBalance);
+              const doesAmountExceedsTokenBalance = tokenBalanceRemainder.isNegative();
+              const newMaxAmountIncludingFee =
+                isTokenEther && !doesAmountExceedsTokenBalance
```
amonsosanz: `isTokenEther` can now be removed

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -188,21 +133,25 @@ const BridgeConfirmation: FC = () => {
...
-  }, []);
+      connectedProvider.data.provider.on("block", onBlock);
 
-  useEffect(() => {
-    // Start estimating gas
-    startGasPolling();
-    return stopGasPolling;
-  }, [startGasPolling, stopGasPolling]);
+      return () => {
+        connectedProvider.data.provider.off("block", onBlock);
```
amonsosanz: Why pass here the `onBlock` listener?

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -60,13 +56,9 @@ const BridgeConfirmation: FC = () => {
...
   const { approve, isContractAllowedToSpendToken, getErc20TokenBalance, tokens } =
     useTokensContext();
-  const [isFadeVisible, setIsFadeVisible] = useState(true);
   const [isBridgeInProgress, setIsBridgeInProgress] = useState(false);
   const [tokenBalance, setTokenBalance] = useState<BigNumber>();
-  const [maxPossibleAmountConsideringFee, setMaxPossibleAmountConsideringFee] =
-    useState<BigNumber>();
-  const [shouldUpdateOnScreenAmount, setShouldUpdateOnScreenAmount] = useState(false);
-  const [shouldUpdateOnScreenFee, setShouldUpdateOnScreenFee] = useState(false);
+  const [maxAmountIncludingFee, setMaxAmountIncludingFee] = useState<BigNumber>();
```
amonsosanz: I preferred the name `maxAmountConsideringFee` over `maxAmountIncludingFee` since this value does not "include" the fee, it's indeed the max balance subtracting the fee. Agreed on removing the "possible" part.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -77,102 +69,55 @@ const BridgeConfirmation: FC = () => {
...
+              const doesAmountExceedsTokenBalance = tokenBalanceRemainder.isNegative();
+              const newMaxAmountIncludingFee =
+                isTokenEther && !doesAmountExceedsTokenBalance
+                  ? amount.sub(tokenBalanceRemainder)
+                  : amount;
+
+              setMaxAmountIncludingFee(newMaxAmountIncludingFee);
+            } else {
+              setMaxAmountIncludingFee(amount);
+            }
```
amonsosanz, line 107: We should keep fading the amount when it changes as the previous implementation did. Otherwise, the amount changes by cut before the fade of the fee starts:

missing-fade-on-amount

This needs to be conditional and occur only when the on-screen amounts (both in token and fiat) are affected by the fee. This is why we had this:

```typescript
const hasOnScreenAmountChanged =
    formattedOldMaxPossibleAmountConsideringFee !==
    formattedNewMaxPossibleAmountConsideringFee;

setShouldUpdateOnScreenAmount(hasOnScreenAmountChanged);
```

All this should be restored.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -77,102 +69,55 @@ const BridgeConfirmation: FC = () => {
...
-    const estimateGas = () => {
-      setEstimatedGas((currentEstimatedGas) =>
-        isAsyncTaskDataAvailable(currentEstimatedGas)
-          ? { status: "reloading", data: currentEstimatedGas.data }
-          : { status: "loading" }
-      );
-      if (formData && connectedProvider.status === "successful" && tokenBalance) {
+  useEffect(() => {
+    if (connectedProvider.status === "successful" && formData && tokenBalance) {
+      const onBlock = () => {
```
amonsosanz: We need to do an initial call to `onBlock` (which by the way I would rename to `updateFee` or `updateGas`), otherwise, the screen will not load until a block is emitted.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: You need to pass the listener you want to remove, if not you'll remove all listeners which can lead to unexpected behaviors in the future.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: That's right! Changed

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: Right, done!

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -77,102 +69,55 @@ const BridgeConfirmation: FC = () => {
...
         const { from, to, token, amount } = formData;
         const destinationAddress = connectedProvider.data.account;
 
-        estimateBridgeGas({
+        console.log("Pasa");
```
amonsosanz: 🤣

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -77,102 +69,55 @@ const BridgeConfirmation: FC = () => {
...
+        setEstimatedGas(
+          estimatedGas.status === "successful"
+            ? { status: "reloading", data: estimatedGas.data }
+            : { status: "loading" }
+        );
```
amonsosanz: This should also be reverted to the original code:

```typescript
setEstimatedGas((oldEstimatedGas) =>
  isAsyncTaskDataAvailable(oldEstimatedGas)
    ? { status: "reloading", data: oldEstimatedGas.data }
    : { status: "loading" }
);
```

We can not have `estimatedGas` as dependency of this effect. And we need to use `isAsyncTaskDataAvailable`.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: Here is where [this change](https://github.com/0xPolygonHermez/zkevm-bridge-ui/pull/223/files#r1003513780) is relevant, to avoid the infinite loop.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -188,21 +133,29 @@ const BridgeConfirmation: FC = () => {
...
+      };
+    }
+  }, [
+    callIfMounted,
+    classes,
+    connectedProvider,
+    estimateBridgeGas,
+    estimatedGas,
+    formData,
+    isBridgeInProgress,
```
amonsosanz: Why is this dep here?

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -188,21 +133,29 @@ const BridgeConfirmation: FC = () => {
...
-    };
-    void estimateGas();
-    estimateGasPollInterval.current = window.setInterval(estimateGas, AUTO_REFRESH_RATE);
-  }, [connectedProvider, formData, tokenBalance, callIfMounted, estimateBridgeGas, notifyError]);
+      };
 
-  const stopGasPolling = useCallback(() => {
-    clearInterval(estimateGasPollInterval.current);
-  }, []);
+      connectedProvider.data.provider.on("block", onBlock);
```
amonsosanz: I think that we need to apply a debouncer for this event. According to my tests, it can be fired several times at once for accumulated blocks.

You can check it:

```diff
diff --git a/src/views/bridge-confirmation/bridge-confirmation.view.tsx b/src/views/bridge-confirmation/bridge-confirmation.view.tsx
index 256e377..69e0a5b 100644
--- a/src/views/bridge-confirmation/bridge-confirmation.view.tsx
+++ b/src/views/bridge-confirmation/bridge-confirmation.view.tsx
@@ -75,7 +75,8 @@ const BridgeConfirmation: FC = () => {

   useEffect(() => {
     if (connectedProvider.status === "successful" && formData && tokenBalance) {
-      const onBlock = () => {
+      const onBlock = (block: number) => {
+        console.log(`Block: ${block}`);
         const { from, to, token, amount } = formData;
         const destinationAddress = connectedProvider.data.account;
```

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -188,21 +133,29 @@ const BridgeConfirmation: FC = () => {
...
-    // Start estimating gas
-    startGasPolling();
-    return stopGasPolling;
-  }, [startGasPolling, stopGasPolling]);
+      if (estimatedGas.status === "pending") {
+        onBlock();
+      }
+
+      return () => {
+        connectedProvider.data.provider.off("block", onBlock);
```
amonsosanz, line 145: This unsubscribe is not working. We are subscribing one time per effect execution and never unsubscribing.

amonsosanz: I've confirmed that the unsubscription works when we do not pass the second param `onBlock`.

amonsosanz: This smells of an [identity comparison issue](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/base-provider.ts#L2186).

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: The thing is that we don't need to call the `onBlock` for the first time, as it already fires with the last mined block when you attach a listener to it. Reverted!

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: Nope, it already fires for the first time with the last mined block, so we don't need extra calls.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: Probably a leftover, removed

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: Right, fixed!

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: Not sure why do you mean by "accumulated blocks", it gets fired for the current block + each new mined block in the chain. Of course time between blocks may vary. Not sure if the debouncing would ver very helpful here as the time between blocks is random, so probably we'll debounce just one request more times than two sequential requests...

The best option would be to cancel the abort the previous request when a new block is available, but it's not possible with `ethers`.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: > We can not have estimatedGas as dependency of this effect.

Why not?

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: Yep, i've added it back again

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: This is not what I experienced

amonsosanz: Check this out (with audio):

https://user-images.githubusercontent.com/15896177/197716161-4bf7f399-98e5-45b0-85d3-cd747f894f3e.mp4

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: This was related to calling `onBlock()` for the initial request. In any case, you already fixed it like this, which is ok:

```typescript
if (estimatedGas.status === "pending") {
  onBlock();
}
```

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: > Not sure why do you mean by "accumulated blocks"

I mean that the callback might not be called in 20 seconds and then called 3 times in half a second with 3 blocks.

> The best option would be to abort the previous request when a new block is available, but it's not possible with `ethers`.

That's why I propose a debouncer.

Check this out:

https://user-images.githubusercontent.com/15896177/197723858-a6bcb1f6-b3bc-4cf7-8aee-e5c839d584ac.mp4

By the way, this is the network activity with only 1 block (5 requests):

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: It's funny because the first time it gets called it retrieves the current block, but when the listener is detached and attached again it doesn't. Check it out:

```ts
void connectedProvider.data.provider.getBlockNumber().then(console.log);
const onBlock = (blockNumber: number) => {
  const { from, to, token, amount } = formData;
  const destinationAddress = connectedProvider.data.account;
  console.log("new block", blockNumber);
  // ...
```

You'll see that the first time `blockNumber` is the last mined block. But this happens only the first time it gets called.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: Comparing the BigNumbers will result in unnecessary fades due to the display precision.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: That's not happening to me, how did you manage to get all that blocks accumulated? Blocks in Ethereum happens each 10s more or less (in Goerli the block time is even higher), so it's impossible to get 3 of them together.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
elias-garcia: In fact, the expected behavior is to get the current block when subscribing to the `"block"` event: https://github.com/ethers-io/ethers.js/issues/1504#issuecomment-827967938

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: This also breaks with negative amounts:

negative-values-fade

All this was there for a reason:

image

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: > It's funny because the first time it gets called it retrieves the current block

Yep, I can confirm that. So the issue I found only occurs when you go back to the form and then back to the confirmation screen. Interesting buggy behavior they have there.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
amonsosanz: It looks like Ethers [queries from their last queried block to the current block](https://github.com/ethers-io/ethers.js/issues/492#issuecomment-491519871) and sometimes the event is fired several times in a row.

We have implemented a debouncer to mitigate this issue.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -77,102 +70,65 @@ const BridgeConfirmation: FC = () => {
...
-                }, msTimeout);
-
-                return oldGas;
+                return amount;
               }
+            })();
+
+            setMaxAmountConsideringFee((oldMaxAmountConsideringFee) => {
+              setShouldAmountFade(!oldMaxAmountConsideringFee?.eq(newMaxAmountConsideringFee));
+              return oldMaxAmountConsideringFee;
```
amonsosanz, line 121: `setShouldAmountFade(!maxAmountConsideringFee?.eq(newMaxAmountConsideringFee))` seems to work for me.

src/hooks/use-debounced-block.ts
```diff
@@ -0,0 +1,61 @@
+import { useCallback, useEffect, useState } from "react";
+
+import { ConnectedProvider } from "src/domain";
+import { AsyncTask } from "src/utils/types";
+
+type DebouncedBlockCallback = (blockNumber: number) => void;
```
amonsosanz: I'd better `import { Listener } from "@ethersproject/providers";`

src/hooks/use-debounced-block.ts
```diff
@@ -0,0 +1,61 @@
...
+const DEBOUNCE_TIME_IN_MS = 750;
+
+const useDebouncedBlock = (connectedProvider: AsyncTask<ConnectedProvider, string>) => {
+  const [listener, setListener] = useState<DebouncedBlockCallback>();
+  const [blockNumber, setBlockNumber] = useState<number>();
+
+  const addBlockListener = useCallback(
+    (cb: DebouncedBlockCallback) => {
+      if (!listener) {
+        setListener(() => cb);
```
amonsosanz: Why this wrap `() => cb`?

amonsosanz: I see... without this we get an infinite loop. This is a smell...

src/hooks/use-debounced-block.ts
```diff
@@ -0,0 +1,61 @@
...
+          connectedProvider.data.provider.off("block", onBlock);
+        }
+      };
+    }
+  }, [blockNumber, connectedProvider, listener]);
+
+  return addBlockListener;
+};
+
+export default useDebouncedBlock;
```
amonsosanz, line 30: I feel this hook is over-complicated, partly because it couples two concepts that should remain abstracted:

* Debouncing
* Managing the subscription to new blocks

I'd like to consider an alternative based on 2 simpler hooks:

A generic, multipurpose, reusable `use-debounce` hook:

```typescript
import { useEffect, useState } from 'react'

export function useDebounce<T>(value: T, delay: number): T {
  const [debouncedValue, setDebouncedValue] = useState<T>(value)

  useEffect(() => {
    const handler = setTimeout(() => {
      setDebouncedValue(value)
    }, delay)

    return () => {
      clearTimeout(handler)
    }
  }, [value, delay])

  return debouncedValue
}
```

A specific `use-debounce-block` that manages the subscription and debounces the blocks using the other hook :

```typescript
import { useEffect, useState } from "react";

import { useDebounce } from "src/hooks/use-debounce";
import { AsyncTask, isAsyncTaskDataAvailable } from "src/utils/types";
import { ConnectedProvider } from "src/domain";

const DEBOUNCE_TIME_IN_MS = 750;

const useDebouncedBlock = (connectedProvider: AsyncTask<ConnectedProvider, string>) => {
  const [blockNumber, setBlockNumber] = useState<number>();

  useEffect(() => {
    if (isAsyncTaskDataAvailable(connectedProvider)) {
      void connectedProvider.data.provider
        .getBlockNumber()
        .then(setBlockNumber)
        .then(() => {
          connectedProvider.data.provider.on("block", setBlockNumber);
        });

      return () => {
        connectedProvider.data.provider.off("block", setBlockNumber);
      };
    }
  }, [connectedProvider]);

  return useDebounce(blockNumber, DEBOUNCE_TIME_IN_MS);
};

export default useDebouncedBlock;
```

It just returns the new debounced block when it arrives. It can be used like this:

```typescript
useEffect(() => {
  if (newBlock) {
    console.log("A new block has been mined:", newBlock);
  }
}, [newBlock]);
```

Which results in:



Do you see any drawbacks to this approach?

Some benefits I see:

* Simpler code to follow/maintain with encapsulated functionality.
* No listeners/callbacks need to be passed to the block debouncer.
* We get a generic debouncer that can be used for anything that requires debouncing (like an input).

src/hooks/use-debounced-block.ts
elias-garcia: It's not a smell, it's how `setState` works. If you pass a callback to the `setState` it will work as `setState((oldValue) => // ...)`, so it's not setting anything if you pass it a callback.

src/constants.ts
```diff
@@ -36,7 +36,9 @@ export const PAGE_SIZE = 25;
 
 export const PENDING_TX_TIMEOUT = 30 * 60 * 1000; // 30min in ms
 
-export const BRIDGE_CALL_GAS_INCREASE_PERCENTAGE = 20; // 20%
+export const BRIDGE_CALL_GAS_LIMIT_INCREASE_PERCENTAGE = 20; // 20%
+
+export const GAS_PRICE_INCREASE_PERCENTAGE = 50; //
```
amonsosanz, line 41: Why not double it as you mentioned Ethers does?

src/hooks/use-debounced-block.ts
elias-garcia: I agree, this is a more `hook` like implementation while i'm more used to node `EventEmitter` like patterns. In this case this will match React architecture better.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -77,132 +64,77 @@ const BridgeConfirmation: FC = () => {
...
+          } else {
+            callIfMounted(() => {
+              notifyError(error);
+            });
+          }
+        });
+    }
+  }, [
+    callIfMounted,
+    classes,
```
amonsosanz: leftover

src/hooks/use-debounced-block.ts
amonsosanz: I don't understand what you mean.

Anyway, this is gone now.

src/views/bridge-confirmation/bridge-confirmation.view.tsx
```diff
@@ -468,49 +401,42 @@ const BridgeConfirmation: FC = () => {
...
-        </Typography>
-        {fiatAmountString && (
-          <Typography
-            className={shouldUpdateOnScreenAmount ? `${classes.fiat} ${fadeClass}` : classes.fiat}
-            type="body2"
-          >
-            {fiatAmountString}
-          </Typography>
-        )}
+        <div>
```
amonsosanz: leftover

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -217,6 +217,7 @@ const BridgeForm: FC<BridgeFormProps> = ({
   );
 
   useEffect(() => {
+    notifyError(new Error("WTF"));
```
amonsosanz: I'd better remove this

src/constants.ts
elias-garcia: Well, the EIP-1559 fee model works quite differently than the legacy fee model so doubling it it's not the same. The old fee model is basically a bid. If you double the price you'll be overpaying a lot for the tx to get into the block.

amonsosanz (approved):

PR #230, opened 2022-10-27
Fix some issues in the activity view

### LINKED ISSUE #226: Fix some issues in the activity view
## Rationale

This issue is focused on fixing the following aspects:

* Fix the pending Bridges length mismatch bug.
* Keep the pending TXs at their original position in the list once the API returns them (and at the top otherwise).
* Make sure that a deposit can not be finalized twice.

### Pending Bridges length mismatch bug

[Here](https://github.com/0xPolygonHermez/zkevm-bridge-ui/blob/develop/src/views/activity/activity.view.tsx#L107) we merge the pending bridges we get from the local storage with a filtered version of the bridges we got from the API. When a bridge is returned from the API, there is no problem, since we filter it out and replace it with its local storage version. But while the bridge is not yet present in the API, we're "synthetically" adding it to the store. This has 2 side effects:

* The actual length of the bridges in the store is greater (by the number of bridges not yet returned by the API) than the value we store in the [lastLoadedItem](https://github.com/0xPolygonHermez/zkevm-bridge-ui/blob/develop/src/views/activity/activity.view.tsx#L145) state (used by the pagination).
* Later when the user asks for the next page, [we compare](https://github.com/0xPolygonHermez/zkevm-bridge-ui/blob/develop/src/views/activity/activity.view.tsx#L134) the length of the bridges in the store (that includes "ghost" pending bridges) with the last `total` value returned by the API to decide if there are new pages that might be requested. This condition will now lie since we synthetically inflated this length.

A better design should avoid merging what we get from the API and what we get from the local storage in the same store.
Instead, the state would always contain what we got from the API without any transformation, and a helper function would take care of computing the merge at presentation time.

### Keep the pending TXs at their original position in the list

Currently, when the users initiate a deposit from the form they are sent to the Activity page. The deposit appears at the top of the list while the transaction is being processed.

In the second step (i.e the claim process), the user clicks the Finalise button and then the deposit suddenly disappears from its current position on the list and is moved to the top. Once the transaction has been processed, the deposit disappears from the top to return to its original position in the list.

These jumps are disconcerting and should not occur. Once a deposit has a position in the list (i.e it's returned from the API) any further changes to its state should not move it from its current position.

### Make sure that a deposit can not be finalized twice

Currently, when the user clicks the Finalise button on a deposit in the Activity list, the Finalise button is disabled for all the deposits in the list so it cannot be clicked again on any deposit. Once the transaction is signed in MetaMask, the Finalise buttons are activated again. The problem is that this happens before the processed deposit status is changed to "processing". This should not happen, since it allows the user to attempt a double claim os the same deposit.

amonsosanz:
Closes #226 

### What does this PR does?

This PR addresses the following:

* Fix the pending Bridges length mismatch bug.
* Keep the pending TXs at their original position in the list once the API returns them (and at the top otherwise).
* Make sure that a deposit can not be finalized twice.

Find a detailed description of each fix in issue #226.

### DIFF
--- a/src/views/activity/activity.view.tsx
+++ b/src/views/activity/activity.view.tsx
@@ -142,7 +142,7 @@ const Activity: FC = () => {
     if (
       env &&
       isAsyncTaskDataAvailable(connectedProvider) &&
-      isAsyncTaskDataAvailable<Bridge[], undefined, true>(apiBridges) &&
+      apiBridges.status === "successful" &&
       apiBridges.data.length < total
     ) {
       setApiBridges({ status: "loading-more-items", data: apiBridges.data });
--- a/src/views/activity/activity.view.tsx
+++ b/src/views/activity/activity.view.tsx
@@ -21,7 +21,7 @@ import {
   isMetaMaskUserRejectedRequestError,
 } from "src/utils/types";
 import { AUTO_REFRESH_RATE, PAGE_SIZE } from "src/constants";
-import { Bridge } from "src/domain";
+import { Bridge, PendingBridge } from "src/domain";
 import useCallIfMounted from "src/hooks/use-call-if-mounted";
 import useIntersection from "src/hooks/use-intersection";
 
@@ -33,7 +33,10 @@ const Activity: FC = () => {
   const { notifyError } = useErrorContext();
   const { openSnackbar } = useUIContext();
   const { tokens } = useTokensContext();
-  const [bridgeList, setBridgeList] = useState<AsyncTask<Bridge[], undefined, true>>({
+  const [apiBridges, setApiBridges] = useState<AsyncTask<Bridge[], undefined, true>>({
+    status: "pending",
+  });
+  const [pendingBridges, setPendingBridges] = useState<AsyncTask<PendingBridge[], undefined>>({
     status: "pending",
   });
   const [displayAll, setDisplayAll] = useState(true);
@@ -64,9 +67,6 @@ const Activity: FC = () => {
         bridge,
       })
         .then(() => {
-          if (isAsyncTaskDataAvailable<Bridge[], undefined, true>(bridgeList)) {
-            processFetchBridgesSuccess(bridgeList.data);
-          }
           openSnackbar({
             type: "success-msg",
             text: "Transaction successfully submitted.",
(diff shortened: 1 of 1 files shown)


src/views/activity/activity.view.tsx
```diff
@@ -86,36 +86,48 @@ const Activity: FC = () => {
...
-      void getPendingBridges(bridges).then((pendingBridges) => {
-        callIfMounted(() => {
-          const filteredApiBridges = bridges.filter((bridge) => {
-            const isBridgePresentInPendingBridges =
-              (bridge.status === "pending" || bridge.status === "on-hold") &&
-              pendingBridges.find(
-                (pendingBridge) => pendingBridge.depositTxHash === bridge.depositTxHash
-              ) !== undefined;
-            return isBridgePresentInPendingBridges ? false : true;
+      setLastLoadedItem(bridges.length);
```
elias-garcia, line 109: What is the difference between `total` and `lastLoadedItem`? Shouldn't they be the same?

src/views/activity/activity.view.tsx
```diff
@@ -86,36 +86,48 @@ const Activity: FC = () => {
           });
         })
         .finally(() => {
-          setAreBridgesDisabled(false);
+          if (isAsyncTaskDataAvailable<Bridge[], undefined, true>(apiBridges)) {
+            getPendingBridges(apiBridges.data)
```
elias-garcia, line 90: Why instead of calling `getPendingBridges` again we just do:

```ts
processFetchBridgesSuccess(apiBridges.data);
```

src/views/activity/activity.view.tsx
amonsosanz: * `total`: How many items there are in the collection
* `lastLoadedItem`: How many items have been loaded so far

src/views/activity/activity.view.tsx
amonsosanz: * Because we don't want to call `setLastLoadedItem` or `setApiBridges` (since nothing has changed API-wise)
* Because we need to listen to the callback of `getPendingBridges` to re-enable the bridges after the new pending bridge is listed and `processFetchBridgesSuccess` returns `void`. This is related to fixing the bug:

> Make sure that a deposit can not be finalized twice.

elias-garcia (approved):

PR #231, opened 2022-10-27
Improve token list balances

### LINKED ISSUE #148: Improve token balances
We need to improve how we display token balances:

1. In the `Home` component, we need to display `0` as balance when we have an ERC-20 token selected in the `from` chain but this token hasn't been wrapped yet in the `to` chain. Currently we are displaying `--`.
2. In the `TokenList` we need to display the `Spinner` when the balance of a token is loading. Currently we are displaying `--`.

elias-garcia:
Closes #148.

### What does this PR does?

This PR improves how token balances are displayed in the app making the `balance` prop of the `Token` domain interface be an `AsyncTask` instead of just a `number`.

It also removes Rinkeby unused tokens, refactors and improves the `BridgeForm` and the old `TokenList` component and fixes some bugs related to the token search input.

## Checklist

review them:


src/views/home/components/token-list/token-list.styles.ts
```diff
@@ -111,7 +74,7 @@ const useListStyles = createUseStyles((theme: Theme) => ({
   },
   list: {
     overflowY: "auto",
-    maxHeight: 270,
+    height: 270,
```
amonsosanz: This height is now adding a bit of extra space that would not accommodate a fifth token:

token-list-height

I asked the design team and they want to fit 6 tokens before showing the scrollbar so this height should be `376`.

src/views/home/components/token-adder/token-adder.view.tsx
```diff
@@ -0,0 +1,48 @@
...
+          Interact carefully with any new or suspicious token
+        </Typography>
+      </div>
+      <TokenInfoTable token={token} className={classes.tokenDetails} />
+      <button className={classes.addTokenButton} onClick={() => onAddToken(token)}>
+        Add {token.name}
+      </button>
+    </div>
+  );
+};
```
amonsosanz, line 35: This layout is broken:

image

src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -1,287 +1,269 @@
...
-                  <XMarkIcon className={classes.closeButtonIcon} />
+  useEffect(() => {
+    if (tokens) {
+      setFilteredTokens(tokens);
+    }
+  }, [tokens]);
+
+  const error =
+    customToken.status === "failed"
+      ? "Token not supported"
```
amonsosanz: I know "design-wise" we are supposed to show only "Token not supported", but shouldn't we provide a more meaningful error like the ones we have in the async task?

* Couldn't retrieve token balance.
* The token contract cannot be called on any network. Please make sure the contract is deployed on a network.

I prefer showing a specific error when possible.

src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -1,287 +1,269 @@
...
+      : searchInputValue.length > 0 && filteredTokens.length === 0
+      ? "No result found"
+      : undefined;
+
+  return (
+    <div>
```
amonsosanz: This should be a fragment.

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -96,99 +78,39 @@ const BridgeForm: FC<BridgeFormProps> = ({
...
-            )
-        );
-    }
-  };
-
-  const onTokenListClosed = () => {
-    setFilteredTokens(undefined);
-    setTokenListCustomToken({ status: "pending" });
-    setTokenListSearchInputValue("");
+  const handleCloseTokenSelector = () => {
```
amonsosanz: Here we now have `onAmountInputChange`, `onTokenDropdownClick`, `handleSelectToken`, and `handleCloseTokenSelector`.

We used to use `on` and in this PR `handle` is introduced. I'm curious about what are the criteria to decide among the prefixes `on` vs. `handle`.

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -217,40 +139,93 @@ const BridgeForm: FC<BridgeFormProps> = ({
...
+            status: "pending",
+          },
+        }))
+      );
+    }
+  }, [defaultTokens, selectedChains]);
+
+  useEffect(() => {
+    // Load the balances of all the tokens of the primary chain (from)
+    const areTokensPending = tokens?.some((tkn) => tkn.balance?.status === "pending");
```
amonsosanz, line 161: I don't understand what this check saves us from. This effect will only run when the tokens or the selected chain change and in those cases, we need to retrieve again all the balances.

Assuming that the effect above is removed, I think that all we need to do in this effect is:

* Set all the tokens as loading
* Request all the balances
* Set again the tokens with their balance

```typescript
useEffect(() => {
  // Load the balances of all the tokens of the primary chain (from)
  if (selectedChains && defaultTokens) {
    const { from } = selectedChains;
    const chainTokens = [...getChainCustomTokens(from), ...defaultTokens];
    setTokens(chainTokens.map((tkn) => ({ ...tkn, balance: { status: "loading" } })));

    Promise.all(
      chainTokens.map((token: Token) =>
        getTokenBalance(token, from)
          .then(
            (balance): Token => ({
              ...token,
              balance: {
                status: "successful",
                data: balance,
              },
            })
          )
          .catch(
            (): Token => ({
              ...token,
              balance: {
                status: "failed",
                error: "Couldn't retrieve token balance",
              },
            })
          )
      )
    )
      .then(setTokens)
      .catch(notifyError); // We won't catch anything here, we can also use void
  }
}, [defaultTokens, selectedChains, getTokenBalance, notifyError]);
```

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -217,40 +139,93 @@ const BridgeForm: FC<BridgeFormProps> = ({
...
+      setTokens(
+        chainTokens.map((token) => ({
+          ...token,
+          balance: {
+            status: "pending",
+          },
+        }))
+      );
+    }
+  }, [defaultTokens, selectedChains]);
```
amonsosanz, line 157: Why is this initialization required? The balance not being present is the same as it being pending for all purposes, right?

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -217,40 +139,93 @@ const BridgeForm: FC<BridgeFormProps> = ({
...
     // Load the balance of the selected token in the secondary network (To)
     if (selectedChains && token) {
       const isTokenEther = token.address === ethersConstants.AddressZero;
+
+      setBalanceTo((currentBalanceTo) =>
+        currentBalanceTo.status === "successful"
+          ? { status: "reloading", data: currentBalanceTo.data }
+          : { status: "loading" }
+      );
+
```
amonsosanz: Some comments here:

 * I don't think it's worth it using the `reloading` state here since the old value won't be shown anywhere.
 * I'd not call `notifyError` below, as we don't notify balance errors in the token list. The error is available in the async task so the view can now decide to show it locally or just show a 0 as it does today.
 * I'd use the function `getTokenBalance` also here.

```typescript
useEffect(() => {
  // Load the balance of the selected token in the secondary network (To)
  if (selectedChains && token) {
    setBalanceTo({ status: "loading" });

    getTokenBalance(token, selectedChains.to)
      .then((balance) =>
        callIfMounted(() => {
          setBalanceTo({ status: "successful", data: balance });
        })
      )
      .catch(() => {
        callIfMounted(() => {
          setBalanceTo({ status: "failed", error: "Couldn't retrieve token balance" });
        });
      });
  }
}, [callIfMounted, getTokenBalance, selectedChains, token]);
```

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -332,9 +300,11 @@ const BridgeForm: FC<BridgeFormProps> = ({
           </div>
           <div className={classes.rightBox}>
             <Typography type="body2">Balance</Typography>
-            <Typography type="body1">
-              {`${formatTokenAmount(balanceFrom || BigNumber.from(0), token)} ${token.symbol}`}
-            </Typography>
+            <TokenBalance
+              token={{ ...token, balance: balanceFrom }}
```
amonsosanz, line 289: Instead of declaring `balanceFrom` and building here the token, we could also set the token with balance once we get the balances in the effect so the selected token already has the balance.

```diff
-         .then(setTokens)
+        .then((tokensWithBalance) => {
+          setTokens(tokensWithBalance);
+          setToken((selectedToken) =>
+            tokensWithBalance.find((tkn) => tkn.address === selectedToken?.address)
+          );
+        })
```

src/views/home/components/token-info/token-info.view.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <button className={classes.backButton} onClick={onNavigateToTokenList}>
+          <ArrowLeftIcon className={classes.backButtonIcon} />
+        </button>
+        <Icon url={token.logoURI} className={classes.tokenIcon} size={28} />
+        <Typography type="h2">{`${token.name} (${token.symbol})`}</Typography>
+        <button className={classes.closeButton} onClick={onClose}>
+          <XMarkIcon className={classes.closeButtonIcon} />
+        </button>
+      </div>
+      <TokenInfoTable token={token} className={classes.tokenDetails} />
```
amonsosanz: Maybe we should rename also the style as `classes.tokenInfoTable`

src/views/home/components/token-adder/token-adder.view.tsx
```diff
@@ -0,0 +1,48 @@
...
+          <XMarkIcon className={classes.closeButtonIcon} />
+        </button>
+      </div>
+      <div className={classes.disclaimerBox}>
+        <WarningIcon className={classes.disclaimerBoxWarningIcon} />
+        <Typography type="body1" className={classes.disclaimerBoxMessage}>
+          Interact carefully with any new or suspicious token
+        </Typography>
+      </div>
+      <TokenInfoTable token={token} className={classes.tokenDetails} />
```
amonsosanz: Maybe we should rename also the style as `classes.tokenInfoTable`

src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -1,287 +1,269 @@
...
+    if (customToken.status === "successful") {
+      setFilteredTokens([customToken.data]);
+    }
+  }, [customToken]);
 
   useEffect(() => {
     if (inputRef.current) {
       inputRef.current.focus();
     }
   }, []);
```
amonsosanz, line 158: The `inputRef` and this effect can now be removed.

I see though that the autofocus does not work when the modal is initially loaded. It only works when we come back from the token info table view.

src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -1,287 +1,269 @@
...
-                    onClick={() => onSearchInputValueChange("")}
+                    className={classes.addTokenButton}
+                    onClick={() => {
+                      addWrappedToken({ token })
+                        .then((newToken) => {
+                          onNavigateToTokenAdder(newToken);
+                        })
+                        .catch(() => {
+                          onNavigateToTokenAdder(token);
+                        });
```
amonsosanz: We should use `callIfMounted` in these `then` and `catch`.

src/views/home/components/token-info-table/token-info-table.view.tsx
```diff
@@ -10,13 +10,13 @@ import { useEnvContext } from "src/contexts/env.context";
...
 
-interface TokenDetailsProps {
+interface TokenInfoTableProps {
   token: Token;
   className?: string;
 }
 
-const TokenDetails: FC<TokenDetailsProps> = ({ token, className }) => {
-  const classes = useTokenDetailsStyles();
+const TokenInfoTable: FC<TokenInfoTableProps> = ({ token, className }) => {
```
amonsosanz, line 18: Technically this is not a table. I prefer naming this component `TokenDetails` or `TokenInfo`.

src/views/home/components/token-selector/token-selector.view.tsx
```diff
@@ -0,0 +1,156 @@
...
+      type: "token-info",
+      token,
+    });
+  };
+
+  const handleNavigateToTokenList = () => {
+    setScreen({
+      type: "token-list",
+    });
+  };
```
amonsosanz: I'd replace the "NavigateTo" part of these 3 handlers with "Show". The current names are confusing since we're not navigating anywhere, but showing a given screen in this same route.

src/views/home/components/token-adder/token-adder.view.tsx
elias-garcia: Right, I created the class but forgot to add it to the component, fixed!

src/views/home/components/token-list/token-list.styles.ts
elias-garcia: Ah I thought that the original height was taking the design into account! Changed to `376` as the design team wants.

src/views/home/components/token-list/token-list.view.tsx
elias-garcia: Well, I don't like to return a fragment as a root component so I've added the corresponding class to this div!

src/views/home/components/token-list/token-list.view.tsx
elias-garcia: I agree! I've fixed this and improved the error msg.

src/views/home/components/bridge-form/bridge-form.view.tsx
elias-garcia: Mmmm good catch! This was a convention we had in the `wallet-ui` and in the `batch-explorer` (you can check it that the `AmountInput` component is still using it). Not sure why we lost it here as it's very descriptive and I really like it. But we are not following it, so I've changed all the `handle...` to `on...`.

This convention is very useful because, for example, it prevents collisions of this type:

Captura de pantalla 2022-11-02 a las 20 40 49

If you rename `handleAddToken` to `onAddToken` will create a collision with the listener being passed from the parent component, so you need to workaround the name to something like `onAddTokenToList` to prevent the collision.

src/views/home/components/token-list/token-list.view.tsx
elias-garcia: Why? f the component unmounts this `onNavigate...` callbacks shouldn't be available right?

src/views/home/components/token-info-table/token-info-table.view.tsx
elias-garcia: We already have a `TokenInfo` component which is basically the container of this one, so renaming this to `TokenDetails` or `TokenInfo` would not help too much...

src/views/home/components/token-info/token-info.view.tsx
elias-garcia: Done!

src/views/home/components/token-adder/token-adder.view.tsx
elias-garcia: Done!

src/views/home/components/token-selector/token-selector.view.tsx
elias-garcia: Well I don't agree with this. In fact, you are navigating through the `TokenSelector` screens, which are `TokenAdder`, `TokenInfo` and `TokenList`. Navigation is a more generic word than just a route change.

src/views/home/components/bridge-form/bridge-form.view.tsx
elias-garcia: You need to set the initial list of the tokens so you are able to see them, at least, in the `TokenList` component without waiting to load all the balances. Taking this into account, it's semantically more correct to add them with a `pending` state in the balance to indicate that your intention is to load them and you can easily do a `switch` later on.

src/views/home/components/bridge-form/bridge-form.view.tsx
elias-garcia: That has the initial problem which is the `Promise.all`. If you do this you won't be able to see a token balance until all of them have finished. With this approach you don't need to wait for all the balances to be able to interact with the list.

src/views/home/components/bridge-form/bridge-form.view.tsx
elias-garcia: Mmm right! I agree with all the bullet points, good improvement here.

src/views/home/components/token-list/token-list.view.tsx
elias-garcia: Yep... I spent some time trying to fix this but I still don't know why it's not working... Will try to give it another try tomorrow.

src/views/home/components/bridge-form/bridge-form.view.tsx
elias-garcia: The con here is that it adds a little bit of complexity and that you still need to wait for the first list of tokens to get loaded... We can revert it anyway to the previous one, I'm not a big fan of any of them...

src/views/home/components/bridge-form/bridge-form.view.tsx
amonsosanz: I'm ok with both prefixes, leave it as you prefer.

src/views/home/components/bridge-form/bridge-form.view.tsx
amonsosanz: Yep, you're right about the `Promise.all`, this is a better solution.

The only improvement I can think of right now is avoiding the initial map and return the loading state after doing the request.

```diff
diff --git a/src/views/home/components/bridge-form/bridge-form.view.tsx b/src/views/home/components/bridge-form/bridge-form.view.tsx
index a28f6b0..d1ed272 100644
--- a/src/views/home/components/bridge-form/bridge-form.view.tsx
+++ b/src/views/home/components/bridge-form/bridge-form.view.tsx
@@ -163,12 +163,7 @@ const BridgeForm: FC<BridgeFormProps> = ({
           ? tokens.map((tkn) => (tkn.address === updatedToken.address ? updatedToken : tkn))
           : undefined;

-      setTokens((oldTokens) =>
-        !oldTokens
-          ? undefined
-          : oldTokens.map((tkn) => ({ ...tkn, balance: { status: "loading" } }))
-      );
-
+      setTokens(
         tokens.map((token: Token) => {
           getTokenBalance(token, selectedChains.from)
             .then((balance): void => {
@@ -197,7 +192,9 @@ const BridgeForm: FC<BridgeFormProps> = ({
                 setTokens((currentTokens) => getUpdatedTokens(currentTokens, updatedToken));
               });
             });
-      });
+          return { ...token, balance: { status: "loading" } };
+        })
+      );
     }
   }, [callIfMounted, defaultTokens, getTokenBalance, selectedChains, tokens]);

```

src/views/home/components/token-list/token-list.view.tsx
amonsosanz: You're right!

src/views/home/components/token-selector/token-selector.view.tsx
amonsosanz: A matter of taste I guess.

src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -1,287 +1,274 @@
...
-                              }}
-                            >
-                              <InfoIcon className={classes.tokenInfoButtonIcon} />
-                            </button>
-                          </div>
-                        )}
+                  <button
+                    className={classes.addTokenButton}
+                    onClick={() => {
+                      addWrappedToken({ token })
```
amonsosanz: We could indeed remove this `addWrappedToken` since now [`getTokenFromAddress` already adds the wrapped token](https://github.com/0xPolygonHermez/zkevm-bridge-ui/blob/develop/src/contexts/tokens.context.tsx#L234).

src/views/home/components/bridge-form/bridge-form.view.tsx
```diff
@@ -217,61 +137,88 @@ const BridgeForm: FC<BridgeFormProps> = ({
...
+          })
+        )
+        .catch(() => {
+          callIfMounted(() => {
+            setBalanceTo({ status: "failed", error: "Couldn't retrieve token balance" });
+          });
+        });
     }
-  }, [selectedChains, account, token, getErc20TokenBalance, notifyError, callIfMounted]);
+  }, [callIfMounted, getTokenBalance, selectedChains, token]);
```
amonsosanz, line 234: We also need to ask for the balance of the From chain since we can no longer rely on the custom token to be part of `tokens` state.

```diff
diff --git a/src/views/home/components/bridge-form/bridge-form.view.tsx b/src/views/home/components/bridge-form/bridge-form.view.tsx
index a28f6b0..3c53f25 100644
--- a/src/views/home/components/bridge-form/bridge-form.view.tsx
+++ b/src/views/home/components/bridge-form/bridge-form.view.tsx
@@ -49,6 +49,9 @@ const BridgeForm: FC<BridgeFormProps> = ({
   const env = useEnvContext();
   const { getErc20TokenBalance, tokens: defaultTokens } = useTokensContext();
   const { connectedProvider } = useProvidersContext();
+  const [balanceFrom, setBalanceFrom] = useState<AsyncTask<BigNumber, string>>({
+    status: "pending",
+  });
   const [balanceTo, setBalanceTo] = useState<AsyncTask<BigNumber, string>>({ status: "pending" });
   const [inputError, setInputError] = useState<string>();
   const [selectedChains, setSelectedChains] = useState<SelectedChains>();
@@ -202,8 +205,22 @@ const BridgeForm: FC<BridgeFormProps> = ({
   }, [callIfMounted, defaultTokens, getTokenBalance, selectedChains, tokens]);

   useEffect(() => {
-    // Load the balance of the selected token in the secondary network (To)
+    // Load the balance of the selected token in both networks
     if (selectedChains && token) {
+      setBalanceFrom({ status: "loading" });
+
+      getTokenBalance(token, selectedChains.from)
+        .then((balance) =>
+          callIfMounted(() => {
+            setBalanceFrom({ status: "successful", data: balance });
+          })
+        )
+        .catch(() => {
+          callIfMounted(() => {
+            setBalanceFrom({ status: "failed", error: "Couldn't retrieve token balance" });
+          });
+        });
+
       setBalanceTo({ status: "loading" });

       getTokenBalance(token, selectedChains.to)
@@ -252,8 +269,6 @@ const BridgeForm: FC<BridgeFormProps> = ({
     );
   }

-  const balanceFrom = tokens?.find((tkn) => tkn.address === token.address)?.balance;
-
   return (
     <form className={classes.form} onSubmit={onFormSubmit}>
       <Card className={classes.card}>
```

amonsosanz: We could also probably use the same effect.

amonsosanz: we can probably use the same effect

src/views/home/components/bridge-form/bridge-form.view.tsx
amonsosanz: This comment has been outdated by [this one](https://github.com/0xPolygonHermez/zkevm-bridge-ui/pull/231#discussion_r1012870831).

src/views/home/components/bridge-form/bridge-form.view.tsx
elias-garcia: Right! Done

src/views/home/components/token-list/token-list.view.tsx
elias-garcia: Right, done!

src/views/home/components/token-selector-header/token-selector-header.view.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC } from "react";
+
+import useTokenSelectorHeaderStyles from "src/views/home/components/token-selector-header/token-selector-header.styles";
+import { ReactComponent as ArrowLeftIcon } from "src/assets/icons/arrow-left.svg";
+import { ReactComponent as XMarkIcon } from "src/assets/icons/xmark.svg";
+import Typography from "src/views/shared/typography/typography.view";
+
+interface TokenSelectorHeaderProps {
+  title: string;
+  onClose?: () => void;
```
amonsosanz, line 10: Why is `onClose` optional? I don't think it has to be optional, but if it's, then the button should render conditionally, otherwise, you see it, click it and it does nothing.

src/views/home/components/token-selector/token-selector.view.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  const onNavigateToTokenAdder = (token: Token) => {
+    setScreen({
+      type: "token-adder",
+      token,
+    });
+  };
+
+  const onNavigateToTokenDetails = (token: Token) => {
```
amonsosanz: Since the type is `"token-info"`, and the component is also `TokenInfo`, I'd rename this to `onNavigateToTokenInfo` both here and in the token-list.

src/views/home/components/token-selector-header/token-selector-header.view.tsx
elias-garcia: Right, forgot to add the conditional render, fixed!

src/views/home/components/token-selector/token-selector.view.tsx
elias-garcia: Done!

amonsosanz (approved):

PR #235, opened 2022-11-09
Fix the permit signature

### LINKED ISSUE #232: Fix the permit signature
## Summary of Bug

Currently, it's not possible to deposit tokens as `UNI` that do not follow the EIP-2612. Uniswap removes the field `version` from the domain separator, so the signature must be computed differently.

We can do as Uniswap and [hardcode a list of tokens](https://github.com/Uniswap/widgets/blob/main/src/hooks/usePermit.ts#L28) per chain, or we can call the contract of the token to guess what type of permit does it implement and then compute the signature accordingly.

To do so we can read the value of [`PERMIT_TYPEHASH`](https://etherscan.io/address/0x1f9840a85d5af5bf1d1762f925bdaddc4201f984#readContract#F3) and if it's not of [Dai type](https://github.com/yashnaman/tokensWithPermitFunctionList), then read [`DOMAIN_TYPEHASH`](https://etherscan.io/address/0x1f9840a85d5af5bf1d1762f925bdaddc4201f984#readContract#F2) or [`EIP712DOMAIN_HASH`](https://etherscan.io/token/0xEEF9f339514298C6A857EfCfC1A762aF84438dEE#readContract#F1) as a fallback for `EIP-2612`.

### Steps to Reproduce

Try depositing UNI from L1 from an account that has not been approved with allowance.

amonsosanz:
Closes #232 

### What does this PR does?

This PR adds support for 3 different implementations of permit:

* DAI
* EIP_2612_STANDARD
* EIP_2612_UNISWAP


src/adapters/ethereum.ts
```diff
@@ -27,26 +32,44 @@ const getConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
...
+const discoverPermit = ({ chain, token }: DiscoverPermitParams): Promise<Permit> => {
   if (token.address === ethersConstants.AddressZero) {
-    return false;
-  }
-  const tokenContractWithPermit = Erc20Permit__factory.connect(token.address, chain.provider);
-  try {
-    return !!(await tokenContractWithPermit.nonces(account));
-  } catch (err) {
-    return false;
+    return Promise.reject();
```
elias-garcia: I would reject with an error even if it's not used to make the code more readable, something like:

```
Promise.reject("Native currency does't require permit")
```

Same with the rejects below.

src/adapters/ethereum.ts
```diff
@@ -27,26 +32,44 @@ const getConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
...
   chain: Chain;
   token: Token;
 }
 
-const isPermitSupported = async ({
-  account,
-  chain,
-  token,
-}: IsPermitSupportedParams): Promise<boolean> => {
+const discoverPermit = ({ chain, token }: DiscoverPermitParams): Promise<Permit> => {
```
elias-garcia: I would call this `getPermitType` or something like that, `discoverPermit` sounds quite weird.

src/adapters/ethereum.ts
```diff
@@ -27,26 +32,44 @@ const getConnectedAccounts = (provider: Web3Provider): Promise<string[]> => {
...
+  const contract = Erc20__factory.connect(token.address, chain.provider);
+  return contract.PERMIT_TYPEHASH().then((permitTypehash) => {
+    switch (permitTypehash) {
+      case DAI_PERMIT_TYPEHASH: {
+        return Permit.DAI;
+      }
+      case EIP_2612_PERMIT_TYPEHASH: {
+        return contract
+          .DOMAIN_TYPEHASH()
+          .catch(() => contract.EIP712DOMAIN_HASH())
```
elias-garcia: I would move the `catch` block below the `then` block to make it easier to follow.

src/constants.ts
```diff
@@ -7,6 +7,18 @@ import { ProofOfEfficiency__factory } from "src/types/contracts/proof-of-efficie
 import { ProviderError } from "src/adapters/error";
 import { Chain, Currency, EthereumChain, EthereumChainId, Token, ZkEVMChain } from "src/domain";
 
+export const DAI_PERMIT_TYPEHASH =
+  "0xea2aa0a1be11a07ed86d755c93467f4f82362b452371d1ba94d1715123511acb";
+
+export const EIP_2612_PERMIT_TYPEHASH =
+  "0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9";
+
+export const EIP_2612_STANDARD_DOMAIN_TYPEHASH =
```
elias-garcia: I would call this just `EIP_2612_DOMAIN_TYPEHASH `, as EIP is an standard itself.

src/domain/index.ts
```diff
@@ -208,3 +208,21 @@ export type Gas =
         gasPrice: BigNumber;
       };
     };
+
+export type TokenSpendPermission =
+  | {
+      type: "allowed";
```
elias-garcia: I would replace this `allowed` discriminator with `none`, which I think it's more aligned with the name of the type.

src/contexts/bridge.context.tsx
```diff
@@ -682,11 +700,17 @@ const BridgeProvider: FC<PropsWithChildren> = (props) => {
...
                   .div(BigNumber.from(100))
                   .mul(BRIDGE_CALL_GAS_LIMIT_INCREASE_PERCENTAGE);
 
-                return gasLimit.add(gasIncrease);
+                const permitGasLimitIncrease = BigNumber.from(
+                  BRIDGE_CALL_PERMIT_GAS_LIMIT_INCREASE
+                );
+
+                return tokenSpendPermission.type === "permit"
+                  ? gasLimit.add(gasLimitIncrease).add(permitGasLimitIncrease)
```
elias-garcia: `add` supports a `BigNumberish` as a param, so we can simplify this a little bit:

```ts
const gasLimitIncrease = gasLimit
  .div(BigNumber.from(100))
  .mul(BRIDGE_CALL_GAS_LIMIT_INCREASE_PERCENTAGE);
const increasedGasLimit = gasLimit.add(gasLimitIncrease)

return tokenSpendPermission.type === "permit"
  ? increasedGasLimit.add(BRIDGE_CALL_PERMIT_GAS_LIMIT_INCREASE)
  : increasedGasLimit
```

src/adapters/ethereum.ts
amonsosanz: Done.

src/adapters/ethereum.ts
amonsosanz: Renamed as `getPermit`.

I named the function `discoverPermit` because it indeed "discovers the `Permit`" doing up to 3 different calls. It's true though that this is an implementation detail.

src/adapters/ethereum.ts
amonsosanz: We can't do that. We need to ask for the `DOMAIN_TYPEHASH()` and fallback to `EIP712DOMAIN_HASH()` (or vice-versa), so we get a hash of either type to later build the `Permit`.

src/constants.ts
amonsosanz: The point is that this "standard" is not always 100% followed, so we end up having 2 flavors within the `EIP_2612` `DOMAIN_TYPEHASH`: _standard_ and _uniswap_.

Does it make sense?

src/domain/index.ts
amonsosanz: This was initially tagged as `non-required` but I changed it because I thought it was confusing:

_"TokenSpendPermission > none"_ is like _"You don't have permission to spend the token"_ while it actually means the opposite.
And since we set this type after calling the `erc20Contract.allowance` method, I thought "allowed" was more clear.

Do you still want me to change it?

src/contexts/bridge.context.tsx
amonsosanz: Yep, much better. Done.

src/adapters/ethereum.ts
elias-garcia: Ahá! You are right, I interpreted this wrong. And wouldn't make sense here to use `Promise.any` instead?

```js
Promise.any([
  contract.DOMAIN_TYPEHASH(),
  contract.EIP712DOMAIN_HASH()
])
```

src/constants.ts
elias-garcia: Yep, in fact that's why I asked. Because, you have the standard type hash, which is `EIP_2612_DOMAIN_TYPEHASH`, and then for any kind of variation (like the UNI token one), you have: `EIP_2612_UNI_DOMAIN_TYPEHASH` as it's a variation of the standard one. But I think that it's redundant to add the `STANDARD` keyword to the standard one as the EIP already indicates that it's the standard one.

src/domain/index.ts
elias-garcia: Mmmm I don't see it that way, because at the end when you see this type, you ask yourself: "Which permission do I need to be able to spend the token?". You probably need an `approval` permission, or maybe a `permit` permission. Or maybe you don't need a permission at all, which should be `none`. But if you ask this yourself and you see `allowed` looks quite weird.

In fact, I think that all the keys of this `Permission` should be nouns: `none`, `approval`, `permit`.

src/adapters/ethereum.ts
amonsosanz: Absolutely! I don't think I've ever used `Promise.any`! Much more descriptive. 👍

Done.

src/constants.ts
amonsosanz: I get the point, but then I think we should remove `EIP_2612` from `EIP_2612_UNISWAP_DOMAIN_TYPEHASH` since this one is not standard.

I mean, if the reasoning is that `EIP_2612_STANDARD_DOMAIN_TYPEHASH` should not include `STANDARD` because `EIP_2612` implies that it follows the standard, then since uniswap does not "completely" follow the standard, we should name it something like `UNISWAP_DOMAIN_TYPEHASH`.

So we end up having: `EIP_2612_DOMAIN_TYPEHASH` and `UNISWAP_DOMAIN_TYPEHASH`.

Agreed?

src/domain/index.ts
amonsosanz: > In fact, I think that all the keys of this `Permission` should be nouns: `none`, `approval`, `permit`.

There you got me. Done.

src/constants.ts
elias-garcia: Mmm makes sense yep, I didn't see that. Okay, taking that into account this last proposal looks good yep!

amonsosanz: Well and of course the type becomes:

```typescript
export enum Permit {
  DAI = "DAI",
  EIP_2612 = "EIP_2612",
  UNISWAP = "UNISWAP",
}
```

amonsosanz: Done

elias-garcia (approved):

PR #237, opened 2022-11-16
New linting proposal

elias-garcia:
### What does this PR does?

This PR formats the code according to https://github.com/0xPolygonHermez/eslint-config-hermez/pull/1. As it's not yet merged and deployed it point to a local tarball to be able to see the consequences of applying the new linting rules. Needs to be replaced with the final version before merging.


amonsosanz:
We're missing an `npm run format`

amonsosanz (approved):

PR #239, opened 2022-12-14
Add outdated pop up

elias-garcia:
### What does this PR does?

This PR refactors the `Policy` component to make it a reusable one and adds a pop-up to alert the user that the network will be outdated in the coming days.

## Checklist

review them:


adrifdez (approved):

src/adapters/env.ts
```diff
@@ -37,6 +39,7 @@ const envToDomain = ({
 }: Env): Promise<domain.Env> => {
   const polygonZkEVMNetworkId = z.number().positive().parse(Number(VITE_POLYGON_ZK_EVM_NETWORK_ID));
   const useFiatExchangeRates = z.boolean().parse(VITE_USE_FIAT_EXCHANGE_RATES === "true");
+  const isNetworkOutdated = z.boolean().parse(VITE_IS_NETWORK_OUTDATED === "true");
```
amonsosanz: This is not safe because it results in `false` for any non-valid value, meaning that whenever the env holds a value different from `true` no error will be shown and the network outdated msg won't be shown.

Instead, we should only succeed with `true` or `false` and provide a meaningful error in any other case.

```diff
diff --git a/src/adapters/env.ts b/src/adapters/env.ts
index 952af38..4327bb6 100644
--- a/src/adapters/env.ts
+++ b/src/adapters/env.ts
@@ -38,8 +38,8 @@ const envToDomain = ({
   VITE_USE_FIAT_EXCHANGE_RATES,
 }: Env): Promise<domain.Env> => {
   const polygonZkEVMNetworkId = z.number().positive().parse(Number(VITE_POLYGON_ZK_EVM_NETWORK_ID));
-  const useFiatExchangeRates = z.boolean().parse(VITE_USE_FIAT_EXCHANGE_RATES === "true");
-  const isNetworkOutdated = z.boolean().parse(VITE_IS_NETWORK_OUTDATED === "true");
+  const useFiatExchangeRates = stringBoolean.parse(VITE_USE_FIAT_EXCHANGE_RATES);
+  const isNetworkOutdated = stringBoolean.parse(VITE_IS_NETWORK_OUTDATED);

   return getChains({
     ethereum: {
@@ -122,6 +122,27 @@ const envParser = StrictSchema<Env, domain.Env>()(
     .transform(envToDomain)
 );

+const stringBoolean = StrictSchema<string, boolean>()(
+  z.string().transform((value, context) => {
+    switch (value) {
+      case "true": {
+        return true;
+      }
+      case "false": {
+        return false;
+      }
+      default: {
+        context.addIssue({
+          code: z.ZodIssueCode.custom,
+          fatal: true,
+          message: "The provided string input can't be parsed as a boolean",
+        });
+        return z.NEVER;
+      }
+    }
+  })
+);
+
 const loadEnv = (): Promise<domain.Env> => {
   const parsedEnv = envParser.parseAsync(import.meta.env);
```

elias-garcia: done!

src/adapters/env.ts
```diff
@@ -29,14 +53,23 @@ const envToDomain = ({
...
   VITE_USE_FIAT_EXCHANGE_RATES,
 }: Env): Promise<domain.Env> => {
   const polygonZkEVMNetworkId = z.number().positive().parse(Number(VITE_POLYGON_ZK_EVM_NETWORK_ID));
-  const useFiatExchangeRates = z.boolean().parse(VITE_USE_FIAT_EXCHANGE_RATES === "true");
+  const useFiatExchangeRates = stringBooleanParser.parse(VITE_USE_FIAT_EXCHANGE_RATES);
+  const bridgeApiUrl = VITE_BRIDGE_API_URL;
+  const outdatedNetwork: domain.Env["outdatedNetwork"] = {
+    message: VITE_OUTDATED_NETWORK_MESSAGE || undefined,
+    title: VITE_OUTDATED_NETWORK_TITLE || undefined,
+    url: VITE_OUTDATED_NETWORK_URL || undefined,
```
amonsosanz: These variables are already `string | undefined`

.env.example
```diff
@@ -20,3 +20,8 @@ VITE_USE_FIAT_EXCHANGE_RATES=true
 VITE_FIAT_EXCHANGE_RATES_API_URL=https://api.exchangeratesapi.io/v1/latest
 VITE_FIAT_EXCHANGE_RATES_API_KEY=
 VITE_FIAT_EXCHANGE_RATES_ETHEREUM_USDC_ADDRESS=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
+
+# MISC
```
amonsosanz, line 24: I'd add a comment here:

`# Both the title and the message are required for the popup to be displayed.`

amonsosanz: In fact, I also consider making the title optional and that only the message is required to display the popup.

src/adapters/env.ts
elias-garcia: Right, I've removed this check from here.

.env.example
elias-garcia: Yeah, it looks good as well without the title. I've marked it as optional.

amonsosanz: Then the message could be:

`# VITE_OUTDATED_NETWORK_MESSAGE is required for the popup to be displayed.`

amonsosanz: nah I don't think the comment is required now

amonsosanz (approved):

PR #243, opened 2023-01-09
Improve deployment and docs

amonsosanz:
### What does this PR does?

* Add `.dockerignore` file to avoid polluting the docker context
* Avoid adding the non-present outdated network modal optional env vars
* Improve docker cache usage for the npm install layer
* Improve docs by adding a docker run example command


Dockerfile
```diff
@@ -4,8 +4,12 @@ RUN apk add --update nodejs npm
 
 WORKDIR /app
 
-COPY . .
+COPY package.json package-lock.json ./
+COPY scripts ./scripts
+COPY abis ./abis
```
adrifdez, line 9: Why this if we are using dockerignore and below you do copy the whole directory?

amonsosanz: By doing this, the step `npm install` can be cached on a layer and won't be rebuilt as long as the files required by the installation process are not modified (package.json, package-lock.json, scripts, and abis).

In other words, you can modify the source code and the step `npm install` won't be rebuilt.

.dockerignore
```diff
@@ -0,0 +1,5 @@
+.env
+.vscode
```
elias-garcia: We don't have a `.vscode` folder in this repo.

Dockerfile
elias-garcia: This was needed yep.

README.md
```diff
@@ -65,6 +65,24 @@ The Docker image won't build the UI until you run it, in order to be able to use
...
+-e ETHEREUM_BRIDGE_CONTRACT_ADDRESS=0x0165878A594ca255338adfa4d48449f69242Eb8F \
+-e ETHEREUM_PROOF_OF_EFFICIENCY_CONTRACT_ADDRESS=0x8dA3b8020401851438eEe8bB434c57b54999935c \
+-e POLYGON_ZK_EVM_RPC_URL=http://localhost:8123 \
+-e POLYGON_ZK_EVM_EXPLORER_URL=http://localhost:4000 \
+-e POLYGON_ZK_EVM_BRIDGE_CONTRACT_ADDRESS=0x9d98deabc42dd696deb9e40b4f1cab7ddbf55988 \
+-e POLYGON_ZK_EVM_NETWORK_ID=1 \
+-e BRIDGE_API_URL=http://localhost:8080 \
+-e USE_FIAT_EXCHANGE_RATES=false \
+-e SHOW_OUTDATED_NETWORK_MODAL=false \
+-p 127.0.0.1:8080:80/tcp --name zkevm-bridge-ui -t zkevm-bridge-ui:local
```
elias-garcia: Why adding the `--name` and `-t` arguments? If we just want to run the image this could be omitted and should be up to each user to tweak them to their needs.

With the port, we should be able to just do: `-p 80:8080`.

adrifdez (approved):

.dockerignore
amonsosanz: I do have it, it's generated by the IDE

amonsosanz: It's in `.gitignore` also

elias-garcia: Yep, `.gitignore` file tries to cover more cases as it prevents this kind of folders to be pushed to the repo. As the app is going to be deployed from a clean repo clone, I don't see necessary to copy all the files from the `.gitignore` here as well, specially those which don't interfere with other things like IDE settings, DS_Store files etc.

README.md
amonsosanz: > Why adding the `--name` and `-t` arguments?

Well, it's just an example command, I think that the reachest the better. A user with little experience in docker can benefit from these example names and choose the ones he wants.

> With the port, we should be able to just do: `-p 80:8080`.

Yep, but if I understand this correctly that would [expose the UI to any machine in the network](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose).

> Note that ports which are not bound to the host (i.e., `-p 80:80` instead of `-p 127.0.0.1:80:80`) will be accessible from the outside.

But sure, we can change this if you prefer it since it might be the desired behavior.

.dockerignore
amonsosanz: Yep, it will never be committed but we have the folder locally in development mode. It's not very relevant in any case.

And you could argue the same with `node_modules` and `dist`, they will never be committed.

README.md
elias-garcia: > > Why adding the `--name` and `-t` arguments?
> 
> Well, it's just an example command, I think that the reachest the better. A user with little experience in docker can benefit from these example names and choose the ones he wants.
> 
> > With the port, we should be able to just do: `-p 80:8080`.
> 
> Yep, but if I understand this correctly that would [expose the UI to any machine in the network](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose).
> 
> > Note that ports which are not bound to the host (i.e., `-p 80:80` instead of `-p 127.0.0.1:80:80`) will be accessible from the outside.
> 
> But sure, we can change this if you prefer it since it might be the desired behavior.

> A user with little experience in docker can benefit from these example names and choose the ones he wants.

Probably a user without experience won't even know what the `name` or `t` parameters mean. I think that we should provide just the minimal command to run the app. Probably a user who needs a name in the service won't even run it alone, he will run it inside a `docker-compose` or similar. And same with the `t` param, the app will be running but why would a regular user want to interact with the container through a terminal?

> Yep, but if I understand this correctly that would [expose the UI to any machine in the network](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose).

Yep, in this case we would like to have access from the outside, as less experimented users won't be aware of this. Also, all docker images examples use this pattern, so I'd go with it.

.dockerignore
elias-garcia: Yep, probably we didn't create this file in the past because it's not needed for our deployments and this files won't never be included in our deployed instances of the bridge ui. Then, if we are going to have this file I'd include at least all the files in the `.gitignore` to be consistent with our deployment.

README.md
amonsosanz: > why would a regular user want to interact with the container through a terminal?

I don't get this.

> I think that we should provide just the minimal command to run the app.

I don't agree. Documentation and examples are better to be verbose.

We could argue long about the names/tags. I think most of this is a matter of taste, but ok, let's use yours. I don't think it's very relevant.

amonsosanz: > why would a regular user want to interact with the container through a terminal?

https://docs.docker.com/engine/reference/run/#foreground

> -t: Allocate a pseudo-tty

🤦 now I get your comment! I thought the `-t` was to specify the image!

amonsosanz: Done!

.dockerignore
amonsosanz: Done!

elias-garcia (approved):

PR #244, opened 2023-01-16
Remove circular dependencies

amonsosanz:
### What does this PR does?

This PR removes the current circular dependencies between modules in the project so we can enable the "import/no-cycle" eslint rule.

graph


Dockerfile
```diff
@@ -12,4 +12,6 @@ RUN npm install
 
 COPY . .
 
+WORKDIR /
```
elias-garcia, line 15: Why do we need to specify the `WORKDIR` at this point? The default one is already the `/` and the ENTRYPOINT is using absolute paths.

src/views/core/router/router.view.tsx
```diff
@@ -1,11 +1,42 @@
...
+import Home from "src/views/home/home.view";
+import Login from "src/views/login/login.view";
+import NetworkError from "src/views/network-error/network-error.view";
+import Settings from "src/views/settings/settings.view";
 import PrivateRoute from "src/views/shared/private-route/private-route.view";
 
+const components: Record<RouteId, { component: ComponentType }> = {
+  activity: {
+    component: Activity,
+  },
```
adrifdez: I think that we can do directly `activity: Activity` the const is called `components`

Dockerfile
amonsosanz: Without this, the image boots at `/app`, but later [this script removes this app folder](https://github.com/0xPolygonHermez/zkevm-bridge-ui/blob/develop/scripts/deploy.sh#L76) resulting in this error when you access the container:

```bash
(unknown) #
sh: getcwd: No such file or directory
```

`getcwd` tries to get the current working directory and if the directory is not found it complains.

src/views/core/router/router.view.tsx
amonsosanz: Absolutely

amonsosanz: Done

elias-garcia (approved):

elias-garcia (approved):

adrifdez (approved):

PR #247, opened 2023-01-20
Display batches remaining to be able to withdraw

### LINKED ISSUE #205: Display batches remaining to be able to withdraw
Currently it's very hard for a user to know when its withdraw is going to be processed. To provide more info about it, we should display in the `BridgeCard` how many batches are left until the batch which includes user tx.

To do this, we need to read the `lastVerifiedBatch` property of the `PoE` smart contract and ask to the L2 node in which batch the user tx is going to be included, so doing a simple subtraction we can now how many batches are left.

The node RPC endpoint will be implemented in https://github.com/0xPolygonHermez/zkevm-node/issues/1281

amonsosanz:
Closes #205

This PR shows the pending batches before the user can finalise the withdraw in the activity screen.

### DIFF
--- a/src/utils/fees.ts
+++ b/src/utils/fees.ts
@@ -1,4 +1,4 @@
-import { TransactionResponse } from "@ethersproject/abstract-provider";
+import { TypeSafeTransactionResponse } from "@ethersproject/abstract-provider";
 import { BigNumber } from "ethers";
 import { Gas } from "src/domain";
 
@@ -14,14 +14,15 @@ export const calculateFee = (gas: Gas): BigNumber => {
 };
 
 export const calculateTransactionResponseFee = (
-  transactionResponse: TransactionResponse
+  transactionResponse: TypeSafeTransactionResponse
 ): BigNumber | undefined => {
   const { gasLimit, gasPrice, maxFeePerGas } = transactionResponse;
-  const gas: Gas | undefined = maxFeePerGas
-    ? { data: { gasLimit, maxFeePerGas }, type: "eip-1559" }
-    : gasPrice
-    ? { data: { gasLimit, gasPrice }, type: "legacy" }
-    : undefined;
+  const gas: Gas | undefined =
+    gasLimit && maxFeePerGas
+      ? { data: { gasLimit, maxFeePerGas }, type: "eip-1559" }
+      : gasLimit && gasPrice
+      ? { data: { gasLimit, gasPrice }, type: "legacy" }
+      : undefined;
 
   return gas && calculateFee(gas);
 };
--- a/src/views/activity/components/bridge-card/bridge-card.view.tsx
+++ b/src/views/activity/components/bridge-card/bridge-card.view.tsx
@@ -1,13 +1,18 @@
-import { FC } from "react";
+import { BigNumber } from "ethers";
+import { FC, useEffect, useState } from "react";
 import { useNavigate } from "react-router-dom";
 
+import { getBatchNumberOfL2Block } from "src/adapters/ethereum";
 import { getCurrency } from "src/adapters/storage";
 import { ReactComponent as BridgeL1Icon } from "src/assets/icons/l1-bridge.svg";
 import { ReactComponent as BridgeL2Icon } from "src/assets/icons/l2-bridge.svg";
-import { Bridge } from "src/domain";
+import { AUTO_REFRESH_RATE } from "src/constants";
+import { AsyncTask, Bridge, Env } from "src/domain";
 import { routes } from "src/routes";
+import { ProofOfEfficiency__factory } from "src/types/contracts/proof-of-efficiency";
 import { formatFiatAmount, formatTokenAmount } from "src/utils/amounts";
 import { getBridgeStatus, getCurrencySymbol } from "src/utils/labels";
+import { isAsyncTaskDataAvailable } from "src/utils/types";
 import useBridgeCardStyles from "src/views/activity/components/bridge-card/bridge-card.styles";
 import Card from "src/views/shared/card/card.view";
 import ErrorMessage from "src/views/shared/error-message/error-message.view";
@@ -16,6 +21,7 @@ import Typography from "src/views/shared/typography/typography.view";
 
 export interface BridgeCardProps {
   bridge: Bridge;
+  env: Env;
   isFinaliseDisabled: boolean;
   networkError: boolean;
   onClaim?: () => void;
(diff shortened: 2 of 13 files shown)


src/views/activity/components/bridge-card/bridge-card.view.tsx
```diff
@@ -55,11 +125,27 @@ const BridgeCard: FC<BridgeCardProps> = ({
...
+    if (
+      isAsyncTaskDataAvailable(lastVerifiedBatch) &&
+      isAsyncTaskDataAvailable(batchNumberOfL2Block)
+    ) {
+      return `Waiting for validity proof. Tx will be confirmed in ${Math.max(
+        batchNumberOfL2Block.data.sub(lastVerifiedBatch.data).toNumber(),
+        0
+      )} batches`;
+    } else if (lastVerifiedBatch.status === "failed" || batchNumberOfL2Block.status === "failed") {
+      return "Waiting for validity proof. It can take between 15 min and 1 hour";
```
amonsosanz: According to [this suggestion](https://0xpolygon.slack.com/archives/C04AD94JAF9/p1674486579845599?thread_ts=1674220096.230739&cid=C04AD94JAF9), this should be:

> Waiting for validity proof. This may take between 15 min and 1 hour

elias-garcia: Let's apply it yep

src/views/activity/components/bridge-card/bridge-card.view.tsx
```diff
@@ -24,14 +30,78 @@ export interface BridgeCardProps {
...
+    status: "pending",
+  });
+  const [batchNumberOfL2Block, setBatchNumberOfL2Block] = useState<AsyncTask<BigNumber, string>>({
+    status: "pending",
+  });
+
+  useEffect(() => {
+    if (status === "initiated" && from.key === "polygon-zkevm") {
+      setBatchNumberOfL2Block({ status: "loading" });
+      getBatchNumberOfL2Block(env.chains[1].provider, depositTxHash)
```
elias-garcia: Instead of sending the `depositTxHash` to later retrieve the `block` in which the tx was included, why we don't just send the `block_num` returned by the API to the function and we just retrieve the batch?

Captura de pantalla 2023-01-24 a las 12 37 14

src/vite-env.d.ts
```diff
@@ -11,8 +21,18 @@ declare global {
   const bridgeVersion: string;
 }
 
+// Module augmentation does not allow changing the types of existing properties (https://github.com/microsoft/TypeScript/issues/36146)
+// but the types of Ethers v5 are broken (https://github.com/ethers-io/ethers.js/issues/2325#issuecomment-977620740), so we provide here
+// a typesafe version of the types and functions that we use in this project
+
+declare module "@ethersproject/abstract-provider" {
+  type TypeSafeTransactionResponse = Nullable<TransactionResponse>;
```
elias-garcia, line 29: Mmmm why setting all the properties of `TransactionResponse` and `TransactionReceipt` to `type | null`? This is not true, only some properties can be `null`.

src/utils/fees.ts
```diff
@@ -14,14 +14,15 @@ export const calculateFee = (gas: Gas): BigNumber => {
...
+  transactionResponse: TypeSafeTransactionResponse
 ): BigNumber | undefined => {
   const { gasLimit, gasPrice, maxFeePerGas } = transactionResponse;
-  const gas: Gas | undefined = maxFeePerGas
-    ? { data: { gasLimit, maxFeePerGas }, type: "eip-1559" }
-    : gasPrice
-    ? { data: { gasLimit, gasPrice }, type: "legacy" }
-    : undefined;
+  const gas: Gas | undefined =
+    gasLimit && maxFeePerGas
```
elias-garcia, line 21: For example, we shouldn't check the `gasLimit` prop here as it will always be there, but we are forced to do it because of the new `TypeSafeTransactionResponse` type.

src/views/activity/components/bridge-card/bridge-card.view.tsx
amonsosanz: Done

src/views/activity/components/bridge-card/bridge-card.view.tsx
amonsosanz: Good catch! I did not remember that the block number was there. It wasn't even parsed yet.

Done!

src/vite-env.d.ts
amonsosanz: Yep, but we don't know which ones. The types of Ethers v5 are broken and Richard Moore, the main maintainer of the project, [recommends we assume that any property can be `null`](https://github.com/ethers-io/ethers.js/issues/2325#issuecomment-977620740):



I added this because I (once again) ran into a runtime error. We've fallen into this trap multiple times, so it's best to be safe and (as they recommend) always assume that all properties can be null. It's always better to extra null check than to get runtime errors.

src/utils/fees.ts
amonsosanz: Yep, that's precisely the point!

src/vite-env.d.ts
elias-garcia: Well, he says that because how the typings have being built. But he also says:

> ... any function needs to be assumed to be null *(although, most of them won’t be)*.

Rewriting the types of all the functions that we use from the library is not an option, even for being type safe. Doing this overrides we can get more runtime errors if at some point they decide to return anything different from the functions. I can accept this changes as an exception because I know that they can return `null` and they are important, but we should not override more ethers types as this have more drawbacks than a runtime error.

elias-garcia (approved):

src/vite-env.d.ts
amonsosanz: > Rewriting the types of all the functions that we use from the library is not an option, even for being type safe.

Of course not! We don't need to override the utils for example. I think this only applies to the values returned by the functions that interact with the blockchain. At least there is where I've found this problem.

> Doing this overrides we can get more runtime errors if at some point they decide to return anything different from the functions.

I don't think this is true and if it were it would not be acceptable. They can return whatever they want in future versions and we won't get any runtime errors. The mapped type `Nullable<Type>` only extends the type they declare (whatever it's today or in any future version) to include `null`, so we're forced to null check. I don't see how this can lead to future runtime errors. As I understand it, this only forces us to be more verbose for the benefit of type safety.

elias-garcia: > The mapped type Nullable<Type> only extends the type they declare **(whatever it's today or in any future version)** to include null, so we're forced to null check.

This is not true, as they can change the interface being returned by the function. Overriding this functions, you won't notice if the interface being returned changes or not. You will only notice it if the import fails (the interface returned is removed from the library). But it's very common to see things like `TransactionResponseV2` when transitioning from one version of a library to another. In this scenario, if they keep the `TransactionResponse` as it's right now, you are doing a wrong assumption.

This can be improved using the [`ReturnType`](https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype) util BTW.

But even doing this we can still have more problems. What if the function name changes? You are assuming that the function is going to be there always and will produce another runtime error. Take into account that you are not only overriding the function signature, but you are **augmenting** this module in case that the functions doesn't exist anymore or change their name.

amonsosanz: I think we're safe in the event of the interface not being available (name change) or having a structural change since in these cases we would get an error at compile time:



It's true though that if the function disappears or is renamed, our augmentation would produce a runtime error, but what are the odds of this happening?

While it's technically possible to have a runtime error if the function disappears, I don't think it's a realistic scenario that we have to consider in the real world, precisely because library authors are reluctant to make changes that break backward compatibility.

PR #250, opened 2023-02-02
Fix tx fees

elias-garcia:
### What does this PR does?

This PR:

- Fixes the way in how we calculate the transaction fee of a confirmed transaction.
- Replaces `JsonRpcProvider` with `StaticJsonRpcProvider` to cache some RPC calls like the `eth_chainId`.
- Move the `lastVerifiedBatch` polling to the `Activity` view to prevent duplicating RPC calls.
- Updates browserlist.

## Checklist

review them:

### DIFF
--- a/src/views/bridge-details/bridge-details.view.tsx
+++ b/src/views/bridge-details/bridge-details.view.tsx
@@ -4,6 +4,7 @@ import { Navigate, useNavigate, useParams } from "react-router-dom";
 
 import { isCancelRequestError } from "src/adapters/bridge-api";
 import { parseError } from "src/adapters/error";
+import { getTxFeePaid } from "src/adapters/ethereum";
 import { getCurrency } from "src/adapters/storage";
 import { ReactComponent as NewWindowIcon } from "src/assets/icons/new-window.svg";
 import { FIAT_DISPLAY_PRECISION, getEtherToken } from "src/constants";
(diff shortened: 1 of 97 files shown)


src/views/bridge-details/bridge-details.view.tsx
```diff
@@ -38,23 +38,27 @@ interface Fees {
...
-          })
+      ? Promise.all([
+          bridge.to.provider.getTransactionReceipt(bridge.claimTxHash),
+          bridge.to.provider.getTransaction(bridge.claimTxHash),
+        ])
+          .then(([txReceipt, txResponse]) =>
+            txReceipt && txResponse
+              ? calculateTransactionReceiptFee({ txReceipt, txResponse, type: "legacy" })
+              : undefined
+          )
```
amonsosanz: Isn't this function assuming that step 1 (deposit) occurs in L1 (so it's always using `eip-1559`) and step 2 (claim) occurs in L2 (so it's always using `legacy`)?

I might be missing something but is this code also correct for a withdrawal? i.e step 1 occurs in L2 and step 2 occurs in L1.

elias-garcia: You are right! Would be great if I can differentiate an EIP-1559 tx receipt from a legacy one... Going to take a look to this

elias-garcia: Ah wait! I already know how to do it hehe

amonsosanz (approved):

PR #265, opened 2023-03-20
Improve network labels

elias-garcia:
### What does this PR does?

This PR improves the labels of the networks across the app.

## Checklist

review them:

### DIFF
--- a/src/utils/labels.ts
+++ b/src/utils/labels.ts
@@ -17,16 +17,13 @@ export function getBridgeStatus(status: Bridge["status"]): string {
   }
 }
 
-export function getNetworkName(chain: Chain): string | undefined {
-  switch (chain.chainId) {
-    case EthereumChainId.MAINNET: {
-      return "Ethereum";
-    }
+export function getEthereumNetworkName(chainId: number): string {
+  switch (chainId) {
     case EthereumChainId.GOERLI: {
-      return "Ethereum Goerli";
+      return "Goerli Testnet";
     }
     default: {
-      return undefined;
+      return "Ethereum";
     }
   }
 }
@@ -34,7 +31,7 @@ export function getNetworkName(chain: Chain): string | undefined {
 export function getDeploymentName(chain: Chain): string | undefined {
   switch (chain.chainId) {
     case EthereumChainId.MAINNET: {
-      return "Mainnet";
+      return "Mainnet Beta";
     }
     case EthereumChainId.GOERLI: {
       return "Testnet";
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -14,6 +14,7 @@ import {
   ZkEVMChain,
 } from "src/domain";
 import { ProofOfEfficiency__factory } from "src/types/contracts/proof-of-efficiency";
+import { getEthereumNetworkName } from "src/utils/labels";
 
 export const DAI_PERMIT_TYPEHASH =
   "0xea2aa0a1be11a07ed86d755c93467f4f82362b452371d1ba94d1715123511acb";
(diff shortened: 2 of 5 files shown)


src/utils/labels.ts
```diff
@@ -17,24 +17,21 @@ export function getBridgeStatus(status: Bridge["status"]): string {
...
-    }
+export function getEthereumNetworkName(chainId: number): string {
+  switch (chainId) {
     case EthereumChainId.GOERLI: {
-      return "Ethereum Goerli";
+      return "Goerli Testnet";
     }
     default: {
-      return undefined;
+      return "Ethereum";
```
amonsosanz, line 26: Why remove the explicit match and fallback to Ethereum?  With this change if we call this function passing the chainId of the zkEVM we will render "Ethereum", right? I'd revert this change to keep forcing the caller to choose the fallback value.

elias-garcia: There are some reasons for this change. The first one is that which fallback will you want instead of `Ethereum`? The second one is that this function is called only when creating the `chains`, so at the end you have some possible values and the default one in another file, which is harder to follow.

amonsosanz: > The first one is that which fallback will you want instead of `Ethereum`? 

It's up to the caller. This function should return `undefined` as before when it's not able to match the `chainId`.

> this function is called only when creating the `chains`

Well, this is today. Tomorrow I might call this function with the `chainId` of another testnet and with this implementation, I would get`Ethereum` as result.

> the default one in another file, which is harder to follow

I don't understand this.

---

My point is that it's not for this function to decide what should be the default value. It's a problem of division of responsibilities. It's the caller, with its context, who should decide what value to use when this function is not able to do a match.

This is what I'd change: 

```diff
diff --git a/src/constants.ts b/src/constants.ts
index 29c9591..a4c0321 100644
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -152,7 +152,7 @@ export const getChains = ({
       explorerUrl: ethereum.explorerUrl,
       Icon: EthChainIcon,
       key: "ethereum",
-      name: getEthereumNetworkName(ethereumNetwork.chainId),
+      name: getEthereumNetworkName(ethereumNetwork.chainId) || "Ethereum",
       nativeCurrency: {
         decimals: 18,
         name: "Ether",
diff --git a/src/utils/labels.ts b/src/utils/labels.ts
index b446f84..1e7e165 100644
--- a/src/utils/labels.ts
+++ b/src/utils/labels.ts
@@ -17,14 +17,17 @@ export function getBridgeStatus(status: Bridge["status"]): string {
   }
 }
 
-export function getEthereumNetworkName(chainId: number): string {
+export function getEthereumNetworkName(chainId: number): string | undefined {
   switch (chainId) {
     case EthereumChainId.GOERLI: {
       return "Goerli Testnet";
     }
-    default: {
+    case EthereumChainId.MAINNET: {
       return "Ethereum";
     }
+    default: {
+      return undefined;
+    }
   }
 }
```

elias-garcia: > Well, this is today. Tomorrow I might call this function with the chainId of another testnet and with this implementation, I would get Ethereum as result.

But that's what we want, to have `Ethereum` as a fallback value if you provide a `chainId` not covered in the switch.

> My point is that it's not for this function to decide what should be the default value

How not? It should return it, that's why you have a `default` case in the `switch` statement, to return the default value 😄

> My point is that it's not for this function to decide what should be the default value. It's a problem of division of responsibilities.

If you follow the separation of concerns pattern you would do it like this... The caller code should get a text label when calling this function instead of having to put a default label when the function doesn't return a label. In the scenario your propose both the caller and the callee will be setting a label, which indeed, is not separating concerns (both have the same concern, the label).

amonsosanz: > But that's what we want

Nope, that's not what we want. As a caller, I want to be aware that the function could not find any name so I can decide the proper fallback for my case. If tomorrow we add support for Sepolia we can forget to process that case and show Ethereum to the user. This code is now more error-prone than before.

> How not? It should return it, that's why you have a `default` case in the `switch` statement, to return the default value 😄

Of course! but the default value should be `undefined` to note that no match was found.

> separation of concerns pattern

No, I'm not talking about any pattern, just about the responsibility of selecting a default value.

---

Well, I am not convinced by these arguments and I do not agree with this change. That being said, do what you prefer. This is not worth more time.

amonsosanz (approved):

PR #267, opened 2023-03-23
Remove deposit limits

elias-garcia:
### What does this PR does?

This PR removes the deposit limits. Now we should just display the warning every time a user wants to do a deposit into the zkEVM.

This PR also removes support for WETH.

## Checklist

review them:


src/views/home/components/deposit-limit-modal/deposit-limit-modal.view.tsx
```diff
@@ -4,103 +4,43 @@ import { SOFT_DEPOSIT_LIMIT_EXCEEDED_WORD } from "src/constants";
...
-          }
-        })()}
+        <Card className={classes.card}>
+          <Typography className={classes.title} type="h1">
+            Warning
+          </Typography>
+          <Typography className={classes.warningText} type="body1">
+            You are about to deposit a large amount of tokens into the Mainnet Beta of the zkEVM.
+            For safety reasons, we need to ensure that you understand its risks and the possible
+            consequences of something going wrong.
```
amonsosanz: deposit-limit

This is not "a large amount of tokens" I think we should change this text if we're showing this msg always.

But are we showing always this msg?? for any number of any token?? this is going to piss off users...

elias-garcia: Yep, I was waiting for the legal text. I'll modify it ASAP

src/constants.ts
```diff
@@ -137,6 +73,19 @@ export const ETH_TOKEN_LOGO_URI =
...
+
+export const POLYGON_PRIVACY_POLICY_URL = "https://polygon.technology/privacy-policy";
+
+export const POLYGON_ZKEVM_RISK_DISCLOSURES_URL =
+  "https://wiki.polygon.technology/docs/zkEVM/#polygon-zkevm-risk-disclosures";
+
+export const TOKEN_BLACKLIST = [
+  // WETH
+  "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+  "0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6",
```
amonsosanz, line 86: Does it make sense to blacklist goerli contracts?

src/views/home/components/token-list/token-list.view.tsx
```diff
@@ -83,50 +84,57 @@ export const TokenList: FC<TokenListProps> = ({
     setCustomToken({ status: "pending" });
 
     if (ethersUtils.isAddress(searchTerm) && newFilteredTokens.length === 0) {
-      setCustomToken({ status: "loading" });
+      if (TOKEN_BLACKLIST.includes(searchTerm)) {
+        setCustomToken({
+          error: "We don't support this token by the moment.",
```
amonsosanz: I think "by the moment" is better worded as "at the moment", I would though change this text to: "We currently do not support this token."

src/constants.ts
```diff
@@ -137,6 +73,19 @@ export const ETH_TOKEN_LOGO_URI =
...
 
+export const POLYGON_TERMS_AND_CONDITIONS_URL = "https://polygon.technology/terms-of-use";
+
+export const POLYGON_PRIVACY_POLICY_URL = "https://polygon.technology/privacy-policy";
+
+export const POLYGON_ZKEVM_RISK_DISCLOSURES_URL =
+  "https://wiki.polygon.technology/docs/zkEVM/#polygon-zkevm-risk-disclosures";
+
+export const TOKEN_BLACKLIST = [
+  // WETH
```
amonsosanz, line 84: Blacklisting `WETH` is problematic because it's used by the price oracle in `src/contexts/price-oracle.context.tsx:91` and in `src/views/bridge-details/bridge-details.view.tsx:209` to get the price of ETH.

src/constants.ts
elias-garcia: Yep, it should be the same behavior on both chains

src/constants.ts
```diff
@@ -43,6 +34,8 @@ export const PENDING_TXS_KEY = "pendingTxs";
 
 export const POLICY_CHECK_KEY = "policyCheck";
 
+export const DEPOSIT_WARNING_KEY = "depositWarning";
```
amonsosanz: I'd use `showDepositWarning` as the key to making clear what it's for.

`DEPOSIT_WARNING_KEY` should also change to `SHOW_DEPOSIT_WARNING_KEY`

src/constants.ts
```diff
@@ -43,6 +34,8 @@ export const PENDING_TXS_KEY = "pendingTxs";
 
 export const POLICY_CHECK_KEY = "policyCheck";
```
amonsosanz, line 35: Does this initial policy check still make sense as it is now?

image

If it remains, shouldn't it also link to the terms and conditions, privacy policy, and/or risk disclosure URLs?

src/views/home/home.view.tsx
```diff
@@ -26,40 +22,31 @@ export const Home = (): JSX.Element => {
...
-      if (depositLimits === undefined) {
-        setDepositLimitModal({
-          data: { formData, type: "unknown-limit" },
-          status: "open",
-        });
-      } else {
-        const softLimit = ethersUtils.parseUnits(depositLimits.soft.toString(), token.decimals);
-        const hardLimit = ethersUtils.parseUnits(depositLimits.hard.toString(), token.decimals);
+  const onCheckShowDepositWarningAndSubmitForm = (formData: FormData) => {
+    const isDepositWarningHidden = getIsDepositWarningHidden();
```
amonsosanz: This PR introduces several references to "[is/getIs/setIs]DepositWarningHidden". This wording is misleading since this does not mean it's "hidden" but "dismissed by the user".

I'd replace all references to `[is/getIs/setIs]DepositWarningHidden` by `[is/getIs/setIs]DepositWarningDismissed`

image

src/views/shared/external-link/external-link.view.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const ExternalLink: FC<PropsWithChildren<ExternalLinkProps>> = ({ children, href }) => {
+  const classes = useExternalLinkStyles();
+
+  return (
+    <a className={classes.link} href={href} rel="noopener noreferrer" target="_blank">
+      {children}
+    </a>
+  );
+};
```
amonsosanz, line 17: We can also consider replacing these now that we have a component:

image

src/constants.ts
elias-garcia: Yes, we are not going to enable the fiat exchange rates for mainnet by the moment, this is addressed in #270.

src/constants.ts
elias-garcia: Naming it `showDepositWarning` doesn't make sense as its default value is `false` and will became `true` when the user checks the checkbox.

src/views/home/home.view.tsx
elias-garcia: Right, done

src/views/shared/external-link/external-link.view.tsx
elias-garcia: I've updated just one, the other ones have different styles (they are links inside a box) and will take more time to adjust them.

src/constants.ts
elias-garcia: Yes, it will be the same by the moment. I'll create another PR if we need to change this.

src/views/home/components/token-list/token-list.view.tsx
elias-garcia: I prefer the first one, updated.

src/constants.ts
amonsosanz: Right, if the logic is inversed the name should be `dismissedDepositWarning`

amonsosanz (approved):

src/constants.ts
elias-garcia: right, done!

PR #300, opened 2024-08-04
Adjust style theme and name [recommendation]

manfioLP:
TODO:
- [ ] test
- [ ] check for more Polygon usages
- [ ] add theme as env variable, default to purple


manfioLP:
opened by accident

